PR #100, opened 2023-05-15
Batches of Batches

ingalls:
### Context

We currently only submit a single "Record" as part of an event that is submitted to the inference handler. As such the iteration over the Records array is essentially a formality and should not alter runtime speed as generally only a single event will be handled.

This PR introduces concurrent record handling within the inference handler to ensure that Network IO heavy steps (calling the SageMaker API) do not block subsequent records from being inferenced. This PR also increases the Record count for Batch Inference requests from the current 1 to 10.

cc/ @nathanielrindlaub

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -327,11 +327,11 @@
             "integrity": "sha512-Xni35NKzjgMrwevysHTCArtLDpPvye8zV/0E4EyYn43P7/7qvQwPh9BGkHewbMulVntbigmcT7rdX3BNo9wRJg=="
         },
         "node_modules/@aws-sdk/abort-controller": {
-            "version": "3.329.0",
-            "resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-3.329.0.tgz",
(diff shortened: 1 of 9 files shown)


nathanielrindlaub:
Quick update/question after some testing. Could the reason we're seeing so many messages land in the DLQ be because if some of the 10 messages in the batch fails, all messages get moved back to the queue? That behavior would explain what I'm seeing in the logs which is: sometimes some images will successfully get predictions and successfully call createLabels, but if something goes wrong with one of the other images in the batch, e.g.:
```
Error: ModelError: Received server error (503) from model with message "
{
    "code": 503,
    "type": "InternalServerException",
    "message": "Prediction failed"
}
```
Then a little while later even the successful messages seem to be getting retried, which throw the `duplicateLabel` error.

I just pushed up a commit that fixes an unrelated bug and attempts to ignore duplicate label errors when they get returned from the graphql-api, but it doesn't seem like a super reliable fix or efficient approach if what I described above is what's happening.

nathanielrindlaub (approved):
Awesome, all looks great!

PR #109, opened 2023-08-02
DeleteImage Function

ingalls:
### Context

Adds a `deleteImage` function to allow an authorized user to delete a single image and all of it's associated assets

- [x] Delete the `Image` object from mdb
- [x] Delete the `ImageAttempt` object from mdb
- [x] Delete any `ImageError` from mdb
- [x] Delete associated S3 Assets

### Example

```
{
    query:`
        mutation DeleteImage($input: DeleteImageInput!){
            deleteImage(input: $input) {
                message
            }
        }
    `,
    variables:{
        input: {
            imageId: '\project:123'
        }
    }
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "animl-serverless",
-    "version": "1.0.0",
+    "version": "2.0.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "animl-serverless",
-            "version": "1.0.0",
+            "version": "2.0.0",
             "license": "ISC",
             "dependencies": {
                 "@aws-sdk/client-lambda": "^3.300.0",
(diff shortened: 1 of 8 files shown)


nathanielrindlaub:
@ingalls I think that checklist above covers it. The two buckets we'll need to find and remove the image files from are:

- `animl-images-serving-<stage>/small/`
- `animl-images-serving-<stage>/original/`
- `animl-images-serving-<stage>/medium/`
- `animl-images-archive-<stage>/<cameraId>/`*

* although side-note I need to revisit the point of the archive bucket. Originally it was to have some place where it would be easy for a human to find all the image files from a given camera, but now that we're allowing duplicate cameraIds across multiple projects it has the potential to become a bit muddied. It's on my to-do list to think on it some more.

ingalls:
Closing in favour of comments in: https://github.com/tnc-ca-geo/animl-api/pull/128

PR #110, opened 2023-08-02
Batch.imageErrors Property

ingalls:
### Context

This PR Implements the `Batch.imageError` property, it also seperates the `BatchError` and `ImageError` model from the user authentication code, making it possible to call Model "helper" functions from other parts of the codebase that don't have access to the user object.

cc/ @nathanielrindlaub


nathanielrindlaub:
Making a note here that IngestDelete is throwing:

```
    "errorType": "AccessDenied",
    "errorMessage": "User: arn:aws:sts::830244800171:assumed-role/animl-ingest-dev-us-west-2-lambdaRole/IngestDelete-dev is not authorized to perform: cloudformation:ListStacks on resource: arn:aws:cloudformation:us-west-2:830244800171:stack/*/* because no identity-based policy allows the cloudformation:ListStacks action",
```

PR #124, opened 2023-09-06
Batch APIs

ingalls:
### Context

- Remove individual Batch API from the public GraphQL Endpoint
- Move AugmentBatch functionality to it's own function to be shared by a filtered aggregate and single ID lookup
- Add Unit tests using new `BatchModel` abtraction to test the above

cc/ @nathanielrindlaub


ingalls:
Merged into https://github.com/tnc-ca-geo/animl-api/pull/123

PR #128, opened 2023-09-25
List Users API

ingalls:
## List Users

Allow Project Managers to list users that are part of their project

### Example

```
{
        query:`
            query ListUsers($input: QueryUsersInput!){
                users(input: $input) {
                    users {
                        username,
                        email,
                        enabled,
                        status,
                        roles,
                        created,
                        enabled
                    }
                }
            }
        `,
        variables:{
            input: {
                filter: 'ingalls'
            }
        }
    }
```

## UpdateUser

- Modify roles by simply submitting a roles array with **ALL** the desired roles for the user
- Remove a user from a project by submitting an empty roles array
- Add a user to a project by submitting at least 1 role

### Example

```
{
            query:`
                mutation UpdateUser($input: UpdateUserInput!){
                    updateUser(input: $input) {
                        message
                    }
                }
            `,
            variables:{
                input: {
                    username: 'ingalls@developmentseed.org',
                    roles: ['manager', 'member']
                }
            }
        }
```

## CreateUser

Allow a project admin to create a new user entry and invite them as a given role to the project

### Example

```
{
            query:`
                mutation createUser($input: CreateUserInput!){
                    createUser(input: $input) {
                        message
                    }
                }
            `,
            variables:{
                input: {
                    username: 'test3@ingalls.ca',
                    roles: ['manager', 'member']
                }
            }
        }
```

## DeleteImages

Adds a `deleteImages` function to allow an authorized user to delete multiple images. Under the hood this calls the single deleteImage function in parallel

### Example

```
{
    query:`
        mutation DeleteImages($input: DeleteImagesInput!){
            deleteImages(input: $input) {
                message
            }
        }
    `,
    variables:{
        input: {
            imageIds: ['project:123']
        }
    }
}
```


ingalls:
```
{
        query:`
            query ListUsers($input: QueryUsersInput!){
                users(input: $input) {
                    users {
                        username,
                        email,
                        enabled,
                        status,
                        role,
                        created,
                        enabled
                    }
                }
            }
        `,
        variables:{
            input: {
                filter: 'ingalls'
            }
        }
    }
```

src/api/db/models/Image.js
```diff
@@ -96,6 +96,32 @@ export class ImageModel {
...
+    try {
+      const s3 = new S3.S3Client({ region: process.env.AWS_DEFAULT_REGION });
+
+      // Ensure Image is part of a project that the user has access to
+      await ImageModel.queryById(input.imageId, context);
+
+      await Promise.all(['medium', 'original', 'small'].map((size) => {
+        return s3.send(new S3.DeleteObjectCommand({
+          Bucket: `animl-images-serving-${process.env.STAGE}`,
+          Key: `${size}/${input.imageId}.jpg`
```
nathanielrindlaub, line 109: Does this work? That doesn't look like it matches the serving bucket Key template (see [image-ingest](https://github.com/tnc-ca-geo/animl-ingest/blob/c75036746ca94b7c503d34cf62cc126eb4d7a9d1/ingest-image/task.js#L267)). We also in theory support .pngs so probably don't want to hard code the file extensions.

src/api/db/models/Image.js
```diff
@@ -96,6 +96,32 @@ export class ImageModel {
     }
   }
 
+  static async deleteImage(input, context) {
```
nathanielrindlaub, line 99: We'll want to support deleting multiple images at once as well.

src/api/db/models/User.js
```diff
@@ -0,0 +1,83 @@
...
+        acc.push(...cur);
+        return acc;
+      }, []).map((user) => {
+        const meta = user.Attributes.reduce((acc, cur) => {
+          acc[cur.Name] = cur.Value;
+          return acc;
+        }, {});
+
+        return {
+          role: user.role,
```
nathanielrindlaub, line 49: Just flagging that currently users can have multiple roles within a single project (e.g. they can be both a 'manager' and 'observer'). I'm just doing a quick scan and haven't read the code here closely enough to confirm whether or not you support this.

nathanielrindlaub (commented):
Looks solid so far @ingalls! I just did a quick pass and added a few comments.

src/api/db/models/User.js
ingalls: It previously supported this but would return two distinct entries for each "role" a user was part of. I've improved this behavior in a subsequent commit to instead return a roles array and perform a deduplication/aggregation on the project level.

src/api/db/models/Image.js
```diff
@@ -100,6 +100,53 @@ export class ImageModel {
...
+      throw new ApolloError(err);
+    }
+  }
+
+  static async deleteImage(input, context) {
+    try {
+      const s3 = new S3.S3Client({ region: process.env.AWS_DEFAULT_REGION });
+
+      // Ensure Image is part of a project that the user has access to
+      const image = await ImageModel.queryById(input.imageId, context);
```
nathanielrindlaub, line 129: Does this actually stop the process if no image is found?

src/api/db/models/Image.js
```diff
@@ -643,6 +690,16 @@ export default class AuthedImageModel {
...
+  async deleteImage(input, context) {
+    if (!hasRole(this.user, WRITE_OBJECTS_ROLES)) throw new ForbiddenError;
+    return await ImageModel.deleteImage(input, context);
+  }
+
+  async deleteImages(input, context) {
+    if (!hasRole(this.user, WRITE_OBJECTS_ROLES)) throw new ForbiddenError;
+    return await ImageModel.deleteImages(input, context);
+  }
+
```
nathanielrindlaub, line 702: I think we need a new role for deleting images, and only managers should be allowed to do it.

Also, why not just keep it to one function (`deleteImages`) that accepts an array of IDs rather than two functions? I'm imagining the client could just pass in an array with one image ID if it only wants to delete a single image. No strong feelings if there's a reason to have both functions.

nathanielrindlaub: Just read some more code and I can see the value in splitting them out and b/c it's nice being able to call `deleteImage` internally, but I'm not sure we need to expose it to clients? Not sure it matters one way or another either. I don't have strong feelings.

src/api/db/models/Image.js
```diff
@@ -100,6 +100,53 @@ export class ImageModel {
...
 
+  static async deleteImages(input, context) {
+    try {
+      const res = await Promise.allSettled(input.imageIds.map((imageId) => {
+        return this.deleteImage({ imageId }, context);
+      }));
+
+      const errors = res
+        .filter((r) => { return r.status === 'rejected'; })
+        .map((r) => { return r.reason; }); // Will always be an ApolloError
```
nathanielrindlaub, line 111: What does that comment `// Will always be an ApolloError` mean? Is that because if anything goes wrong in the `deleteImage` try block, it will be caught and an `ApolloError` will be thrown?

src/api/type-defs/payloads/StandardErrorPayload.js
nathanielrindlaub, line 1: @ingalls this might be easier to kick to slack or our next huddle, but I'm curious where we're at with the goal of throwing ApolloErrors and letting Apollo handle the payload formatting. Are you thinking we wait until upgrading to V4 to do this?

src/api/type-defs/payloads/UsersConnection.js
nathanielrindlaub, line 1: I had only been using the word `Connection` in payloads if it was a paginated query and contained info about the page. I don't think that's true of the users query so for consistency maybe let's call it something else like `UsersPayload`?

src/api/type-defs/payloads/StandardErrorPayload.js
ingalls: Per voice we will keep the errors array for now.

src/api/db/models/Image.js
ingalls: - New Delete Objects role has been created
- The `deleteImage` external API has been removed in favour of the single `deleteImages`

src/api/type-defs/payloads/UsersConnection.js
ingalls: Renamed.

src/api/db/models/Image.js
ingalls: @nathanielrindlaub Per slack, the `r.reason` itself is always an instance of an ApolloError. I didn't see a straightforward way in the docs to throw multiple errors simultaneously so for the moment I've stayed with the current errors array output.

nathanielrindlaub (commented):
@ingalls I encountered a few bugs that need ironing out, but this looks close.

src/api/db/models/User.js
```diff
@@ -0,0 +1,227 @@
...
+   * @param {object} input
+   * @param {string} input.username Email/Username to create
+   * @param {string[]} input.roles List of roles the user should have within the project
+   * @param {object} context
+   */
+  static async create(input, context) {
+    const cognito = new Cognito.CognitoIdentityProviderClient();
+
+    try {
+      await cognito.send(new Cognito.AdminCreateUserCommand({
```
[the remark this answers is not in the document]
nathanielrindlaub: Both look fixed! TY!

src/api/db/models/User.js
```diff
@@ -0,0 +1,227 @@
...
+
+
+  /**
+   * List Users part of a given group
+   *
+   * @param {object} input
+   * @param {string} input.filter Filter usernames by string
+   * @param {object} context
+   */
+  static async list(input, context) {
```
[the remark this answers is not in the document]
nathanielrindlaub: @ingalls fixed for now by updating the `Cognito.ListUsersInGroupCommand()` limit.

PR #130, opened 2023-09-26
Apollo V4

ingalls:
### Context

Apollo V3 is deprecated Nov 15, 2023. This PR is a very preliminary look at what updating to V4 will look like.


nathanielrindlaub:
For reference when we do this upgrade: https://www.apollographql.com/docs/apollo-server/migration

And migration consideration re. error handling specifically: https://www.apollographql.com/docs/apollo-server/migration/#apolloerror

Issue #135, opened 2023-11-27
Image Comments


ingalls:
### Context

- [ ] Add an `Image.comment` field to Database
```json
[{"comment": "I a a comment", "userId": 123, "created": "<date>"}]
```

- [ ] Update Resolvers to support creating/editing? comment
- [ ] Support filtering by has comment/no comment
- [ ] Stretch Goal: Filter by comment internals

Ref: https://github.com/tnc-ca-geo/animl-frontend/issues/164

PR #137, opened 2023-11-28
Image Comment Threads

ingalls:
### Context

This PR implements new fields and an API for creating Image Comment Threads.
- A new role was introduced to manage who can comment
- Non-Admins can only edit/delete their own comments

### Create Comment

Closes: https://github.com/tnc-ca-geo/animl-api/issues/135

```
{
    query:`
        mutation createImageComment($input: CreateImageCommentInput!){
            createImageComment(input: $input) {
                comment {
                    _id
                    author
                    created
                    comment

                }
            }
        }
    `,
    variables:{
        input: {
            imageId: 'default_project:06374dcde709ff578235abefd9eec683',
            comment: 'Test Comment'
        }
    }
}
```

### Update Comment

```
{
    query:`
        mutation updateImageComment($input: UpdateImageCommentInput!){
            updateImageComment(input: $input) {
                comment {
                    _id
                    author
                    created
                    comment
                }
            }
        }
    `,
    variables:{
        input: {
            imageId: 'default_project:06374dcde709ff578235abefd9eec683',
            id,
            comment: 'I AM UPDATED'
        }
    }
}
```

### Delete Comment

```
{
    query:`
        mutation deleteImageComment($input: DeleteImageCommentInput!){
            deleteImageComment(input: $input) {
                message
            }
        }
    `,
    variables:{
        input: {
            imageId: 'default_project:06374dcde709ff578235abefd9eec683',
            id
        }
    }
}
```


src/api/db/models/Image.js
```diff
@@ -316,6 +322,82 @@ export class ImageModel {
...
+
+      if (comment.author !== context.user['cognito:username'] && !context.user['is_superuser']) {
+        throw new Error('Can only edit your own comments');
+      }
+
+      image.comments = image.comments.filter((c) => { return c._id !== input.id; });
+
+      await image.save();
+
+      return { message: 'Images Deleted' };
```
nathanielrindlaub, line 340: @ingalls I just mentioned this in Slack, but I think it would be a bit cleaner on the front end if all of these mutations returned the updated `image.comments` array.

src/api/db/models/Image.js
```diff
@@ -316,6 +322,82 @@ export class ImageModel {
...
+    }
+  }
+
+  static async createComment(input, context) {
+    const operation = async (input) => {
+      return await retry(async (bail, attempt) => {
+        if (attempt > 1) console.log(`Retrying createComment operation! Try #: ${attempt}`);
+
+        // find images, add comment, and bulk write
+        return await Image.bulkWrite([{
```
nathanielrindlaub, line 377: I don't have a strong opinion on this, but I don't think we need to necessarily use `bulkWrite()` here. The reason why I refactored the `updateObjects()` and `upateLabels()` resolvers to use `bulkWrite()` was to support the ability for the frontend to make calls to the API to perform multpile CRUD operations on `image.objects` at once (see PR #127), which I don't think we need to support for image comments.

So this obviously works, but I think we could also follow the `ImageModel.queryById`, make the update, then `image.save()` pattern if there was any advantage to that.

src/api/db/schemas/Image.js
```diff
@@ -13,6 +14,13 @@ const Schema = mongoose.Schema;
  *    userSetData  - user configured EXIF data
  */
 
+const ImageCommentSchema = new Schema({
+  _id: { type: String, required: true, default: randomUUID },
```
nathanielrindlaub, line 18: Is there a reason to manually set the `_id` here? Unless we want the _ids to be something other than one of Mongoose's randomly generated `ObjectIds`, I think if we leave the _id field out entirely Mongoose will [add it for us](https://mongoosejs.com/docs/guide.html#_id).

nathanielrindlaub (changes requested):

nathanielrindlaub (approved):

PR #139, opened 2023-12-01
Project Labels in Database

ingalls:
### Context

Background Reading: https://github.com/tnc-ca-geo/animl-frontend/issues/124

This PR introduces an explicit array of Project labels on the `Project` model in MongoDB. This has the beneficial effects of

- Cheaply (as in speed) knowing what labels are in use in a project
- Allowing Project admins to add & control labels for human reviewers that are more specific or additional to the labels that a model outputs. IE: `Rat` from the model vs `Rattus rattus frugivorus`

### Actions

- [x] Add a `color` field to all MLModel Categories in Mongo - https://gist.github.com/ingalls/927e2c9958976184496a9d9d19a0af1c
- [x] Update MLModel API to allow the exposure of the color field 
- [x] Create a new `Project.labels` schema to hold currently active labels in the project
- [x] Backfill `Project.labels` schema with the output of the current project label resolver and figure out how to backfill colours?! - https://gist.github.com/ingalls/7b7325688a6fd24e9c49cc0992530043
- [x] When a new Image is inferenced, compare the labels to the labels currently on the project and add new labels as necessary
- [x] Add an API to create new labels on the project for project admins
- [x] Add an API to modify existing labels on the project for project admins
- [x] Add an API to delete existing labels on the project for project admins

### DIFF
--- a/test/ml-inference-mega.test.js
+++ b/test/ml-inference-mega.test.js
@@ -80,7 +80,7 @@ tape('ML-Inference Megadetector', async (t) => {
       mlModelVersion: 'v5.0a',
       bbox: [0.28664860129356384, 0.45518332719802856, 0.5675788521766663, 0.6615734100341797],
       conf: 0.9314358830451965,
-      labelId: 'animal'
+      labelId: 1
     }]);
   } catch (err) {
     t.error(err);
@@ -159,7 +159,7 @@ tape('ML-Inference Megadetector - Batch Image', async (t) => {
       mlModelVersion: 'v5.0a',
       bbox: [0.28664860129356384, 0.45518332719802856, 0.5675788521766663, 0.6615734100341797],
       conf: 0.9314358830451965,
-      labelId: 'animal'
+      labelId: 1
     }]);
   } catch (err) {
     t.error(err);


src/api/db/models/Image.js
```diff
@@ -482,6 +465,85 @@ export class ImageModel {
...
+      return { ok: true };
+    } catch (err) {
+      // if error is uncontrolled, throw new ApolloError
+      console.log(`Image.createInternalLabels() ERROR on image ${input.imageId}: ${err}`);
+      if (err instanceof ApolloError) throw err;
+      throw new ApolloError(err);
+    }
+  }
+
   static async createLabels(input, context) {
```
nathanielrindlaub, line 547: I don't have super strong feelings here, but it might be worth better distinguishing between the act of _creating_ a new `Project.label` and _applying_ a label to an image. In other words maybe this action/method should be called something like `applyLabels` or `createLabelRecords`. 

Not a high priority though. What do you think?

src/api/db/models/Image.js
```diff
@@ -482,6 +465,85 @@ export class ImageModel {
...
+
+        // find image, create label record
+        const image = await ImageModel.queryById(label.imageId, context);
+        if (isLabelDupe(image, label)) throw new DuplicateLabelError();
+
+        const project = await ProjectModel.queryById(image.projectId);
+        const labelRecord = createLabelRecord(label, label.mlModel);
+
+        // Check if Label Exists on Project and if not, add it
+        if (!project.labels.some((l) => { return l._id === labelRecord.labelId; })) {
```
nathanielrindlaub, line 491: Any time I'm doing `_id` equality checks w/ MongoDB _ids I always coerce them to strings first to make sure they're the same type. There's a helper function for this in utils.js, `idMatch()`, if you want to use it.

src/api/db/models/MLModel.js
```diff
@@ -3,6 +3,20 @@ import MLModel from '../schemas/MLModel.js';
 import retry from 'async-retry';
 
 export class MLModelModel {
+  static async queryById(_id) {
```
nathanielrindlaub, line 6: Do we need this? Shouldn't the method below `getMLModels({ _ids: [some_id]})` below work?

src/api/db/models/Project.js
```diff
@@ -8,12 +8,27 @@ import { sortDeps, hasRole, idMatch } from './utils.js';
...
 import retry from 'async-retry';
 import {
+  WRITE_PROJECT_ROLES,
   WRITE_DEPLOYMENTS_ROLES,
   WRITE_VIEWS_ROLES,
   WRITE_AUTOMATION_RULES_ROLES
 } from '../../auth/roles.js';
 
 export class ProjectModel {
+  static async queryById(_id) {
```
nathanielrindlaub, line 18: Similar to the `MLModelModel`, do we need this? Shouldn't `getProjects({ _ids: [some_id]})` work?

src/api/db/models/Project.js
```diff
@@ -435,6 +466,49 @@ export class ProjectModel {
       throw new ApolloError(err);
     }
   }
+
+  static async createLabel(input, context) {
+    try {
+      const project = await this.queryById(context.user['curr_project']);
+
+      if (project.labels.filter((label) => {
+        return label.name === input.name;
```
nathanielrindlaub, line 475: I think we should make this case insensitive too.

src/api/db/schemas/Project.js
```diff
@@ -60,6 +61,13 @@ const DeploymentSchema = new Schema({
   editable: { type: Boolean }
 });
 
+const ProjectLabelSchema = new Schema({
+  _id: { type: String, required: true, default: randomUUID },
```
nathanielrindlaub, line 65: I don't think we need to use `randomUUID` unless you don't like MongoDB's `ObjectIds` for some reason. You don't even have to define the `_id` field in the schema, Mongoose will add it for you if you omit it: https://mongoosejs.com/docs/guide.html#_id

nathanielrindlaub: I think the only reason to not use `ObjectIds` would be if you want to set it to something that has meaning for other reasons, like it needs to be human readable or it's a camera serial number or something.

src/api/db/schemas/shared/index.js
```diff
@@ -27,14 +27,13 @@ const ValidationSchema = new Schema({
...
- *    category - the actual label (e.g. "skunk")
  *    conf - confidence of prediction
  *    bbox - [ymin, xmin, ymax, xmax], normalized (not absolute)
  */
 
 const LabelSchema = new Schema({
-  type: { type: String, enum: ['manual', 'ml'], requried: true },
-  category: { type: String, default: 'none', required: true },
+  type: { type: String, enum: ['manual', 'ml'], requried: true, default: 'manual' },
+  labelId: { type: String, required: true },
```
nathanielrindlaub, line 36: Is there a reason we don't just use the standard `_id` field here?

src/api/type-defs/inputs/UpdateProjectInput.js
```diff
@@ -0,0 +1,6 @@
+export default `
+  input UpdateProjectInput {
+    name: String
+    description: String
+  }
```
nathanielrindlaub, line 5: Is the `updateProject` mutation resolver just used to update the name and description of the project? Is it being used in the frontend?

src/api/db/schemas/Project.js
```diff
@@ -73,7 +81,8 @@ const ProjectSchema = new Schema({
   views: { type: [ViewSchema], required: true },
   cameraConfigs: { type: [CameraConfigSchema] },
   availableMLModels: { type: [{ type: String, ref: 'MLModel' }] },
-  automationRules: { type: [AutomationRuleSchema] }
+  automationRules: { type: [AutomationRuleSchema] },
+  labels: { type: [ProjectLabelSchema], default: [], required: true }
```
nathanielrindlaub, line 85: @ingalls, can you add 2 default labels here? One for 'empty' (`color: '#8D8D8D'`) and one for 'unknown' (`color: '#E93D82'`)?

nathanielrindlaub (commented):
@ingalls this is looking super solid. Just a few small questions and suggestions!

src/api/db/models/Image.js
nathanielrindlaub: Hold off for now - maybe cut a separate PR for this

Issue #148, opened 2024-02-12
Async task state tracking


nathanielrindlaub:
We now have a growing number of potentially long-running tasks that Lambdas are not well suited to support, especially if the user is expecting a synchronous response. These include:

- [x] Bulk upload
- [ ] [Deleting large numbers of images](https://github.com/tnc-ca-geo/animl-frontend/issues/227)
- [ ] Deleting large numbers of labels (high priority, very expensive & hard to optimize with `bulkWrite`)
- [ ] Merging large numbers of labels
- [ ] Creating large numbers of labels
- [ ] [Updating camera serial numbers](https://github.com/tnc-ca-geo/animl-frontend/issues/228)
- [x] Getting View stats
- [x] Exporting annotations
- [x] Updating deployments

@ingalls recommends that we consider creating a consistent pattern for tracking these async tasks in the DB (creating a collection or collections that we update when the state of one of these processes changes and a consistent query pattern for accessing them). I think it's a great idea... right now we have 2 entirely different ways of checking the bulk upload state and the annotation state, and for most of the others we haven't yet implemented spinning the tasks off on separate infrastructure; instead, we only support them at pretty low thresholds. 

For the tasks we haven't yet broken out to run async, we have a couple options:
1. Using AWS Batch (same process as batch upload) w/ Fargate. This would be more expensive and have a much longer cold-start, but it would mean that we wouldn’t be limited to Lambda’s 15 min limit
2. Create SQSs + Lambda workers to pull messages off and process them in separate Lambdas. Nick's main concern here is how do we make the UX make sense for tasks that run longer than 15 mins (and would require re-prompting/initiating by the user)

postfalk:
I think that is a good idea. I have a similar design with my Lens to Salesforce workflow.

Falk



Falk Schuetzenmeister

Engineering Lead

The Nature Conservancy in California
***@***.***

(510) 604-8010


[cid:2ed45c50-8c24-4344-a0cf-581e1a0a6237]<https://www.nature.org/en-us/about-us/where-we-work/united-states/california/?vu=california>





















________________________________
From: Natty Rindlaub ***@***.***>
Sent: Monday, February 12, 2024 1:22 PM
To: tnc-ca-geo/animl-api ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [tnc-ca-geo/animl-api] Async task state tracking (Issue #148)


[...]

PR #159, opened 2024-03-12
TaskMan

ingalls:
### Context

There is a need for async task management for longer running tasks than are acceptable during an http request/response cycle. 

This PR introduces a framework for task management and status monitoring

- [ ] Allow listing tasks


nathanielrindlaub:
Ref: #148

src/api/db/models/Project.js
```diff
@@ -488,7 +488,9 @@ export class ProjectModel {
       const label = (project.labels || []).filter((p) => { return p._id.toString() === input._id.toString(); })[0];
       if (!label) throw new DeleteLabelError('Label not found on project');
 
-      const count = await ImageModel.countImagesByLabel([input._id], context);
+      const count = await ImageModel.countImagesByLabel([input._id], {
```
nathanielrindlaub, line 491: @ingalls - guessing this is a relic from when you tried to limit the count aggregation? At any rate it doesn't look like that `limit` option is used or parsed on the `countImagesByLabel` function.

src/api/db/schemas/Task.js
```diff
@@ -0,0 +1,27 @@
...
+  },
+  status: {
+    type: String,
+    required: true,
+    enum : ['SUBMITTED', 'RUNNING', 'FAIL', 'COMPLETE'],
+    default: 'SUBMITTED'
+  },
+  created: { type: Date, default: Date.now, required: true },
+  updated: { type: Date, default: Date.now, required: true },
+  output: { type: Object }
```
nathanielrindlaub, line 22: @ingalls - In the case of `getStats`, the complete output will be a JSON object, but this may not be the case for other async tasks (e.g. `export`'s output gets written to S3, so we return a URL to it to the frontend, other async tasks like updating deployments will probably not have an output unless maybe something fails and we decide we want to return an message).

Anyhow I'm wondering what your thoughts are on how we'll handle those cases?

src/api/db/models/Task.js
```diff
@@ -0,0 +1,101 @@
...
+   * @param {Object} input
+   * @param {String} input.limit
+   * @param {String} input.paginatedField
+   * @param {String} input.sortAscending
+   * @param {String} input.next
+   * @param {String} input.previous
+   * @param {Object} context
+   */
+  static async queryByFilter(input, context) {
+    return await MongoPaging.aggregate(Task.collection, {
```
nathanielrindlaub, line 27: I don't have strong feelings here, but returning a paginated response here is probably overkill here unless at some point we want to provide a long list of previously completed tasks, which I can't really think of a great use case for. LMK if you have thoughts @ingalls.

src/api/db/models/Task.js
```diff
@@ -0,0 +1,101 @@
...
+      aggregation: [
+        { '$match': { 'projectId': context.user['curr_project'] } },
+        { '$match': { 'user': context.user.sub } }
```
nathanielrindlaub, line 30: Oh yeah looks like here we are using `user.sub` but in the Task creation we're using `user.username`?

nathanielrindlaub (commented):
Looking great @ingalls - just a few comments in there.

src/api/db/schemas/Task.js
nathanielrindlaub: As discussed on Slack, we'll just return `output: { url: "s3 string" }` in the case of tasks that produce large files.

PR #165, opened 2024-03-21
Export => Task Manager [Breaking Change]

ingalls:
### Context

Exports currently use an S3 JSON file to store a config document to report export status & results, This PR removes all export infrastructure
- S3 Config File Generation and Retrieval
- Export SQS queue
- Export Function
And merges the export functioanlity into the new Tasking infrastructure for reduced code complexity. Export task status will now be managed in mongo via the `Task` object and status can be tracked using the Get Task endpoint.

This PR will require frontend updates and is breaking


nathanielrindlaub (approved):
Looks great! Ready to merge.

PR #182, opened 2024-04-16
Fixing Local development

jue-henry:
Trying to fix local development for animl-api. Seems like we need to move our animl-api lambda from being ran with an ecr image and instead using the native lambda execution environment now that it supports node20. Also updates serverless offline to run using the right profile and disable the authorizer.

Original changes from #178  and #180

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "description": "Lambda-based graphQL API for camera trap data management platform",
     "main": "src/handler.js",
     "scripts": {
-        "start": "serverless offline --noAuth",
+        "start": "AWS_PROFILE=animl serverless offline --noAuth",
         "deploy-dev": "sls deploy --stage dev",
         "deploy-prod": "sls deploy --stage prod",
         "seed-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "description": "Lambda-based graphQL API for camera trap data management platform",
     "main": "src/handler.js",
     "scripts": {
-        "start": "serverless offline",
+        "start": "serverless offline --noAuth",
         "deploy-dev": "sls deploy --stage dev",
         "deploy-prod": "sls deploy --stage prod",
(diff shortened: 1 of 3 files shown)


serverless.yml
```diff
@@ -148,9 +142,8 @@ functions:
     memorySize: 3008
     timeout: 30
   inference:
-    image:
-      name: base
-      command: src/ml/handler.inference
+    handler: src/api/handler.server
```
nathanielrindlaub, line 145: this should be: `handler: src/ml/handler.inference`

serverless.yml
```diff
@@ -162,15 +155,13 @@ functions:
           functionResponseType: ReportBatchItemFailures
     timeout: 120
   batchinference:
-    image:
-      name: base
-      command: src/ml/handler.inference
+    handler: src/api/handler.server
```
nathanielrindlaub, line 158: This should be: `src/ml/handler.inference`

serverless.yml
```diff
@@ -162,15 +155,13 @@ functions:
...
+    runtime: nodejs20.x
     reservedConcurrency: 8
     timeout: 120
   task:
-    image:
-      name: base
-      command: src/task/handler.handler
+    handler: src/api/handler.server
```
nathanielrindlaub, line 163: this should be: `src/task/handler.handler`

package.json
```diff
@@ -5,7 +5,7 @@
     "description": "Lambda-based graphQL API for camera trap data management platform",
     "main": "src/handler.js",
     "scripts": {
-        "start": "serverless offline",
+        "start": "AWS_PROFILE=animl serverless offline --noAuth",
```
nathanielrindlaub, line 8: @jue-henry, just curious, but do we need `--noAuth` here?

nathanielrindlaub (changes requested):

package.json
jue-henry: We get this warning while running without it
`Warning: Serverless Offline does not support non local authorizers (arn): [object Object]`

I think this option disables the need to include `x-api-key` which actually isn't a huge deal if we test exclusively using the frontend, but this also gets rid of the warning.

nathanielrindlaub: Got it!

nathanielrindlaub (approved):

PR #188, opened 2024-05-29
TypeScript: 1. Initial Tooling

alukach:
## What I'm adding

Add basic initial tooling to support TypeScript within the project, without actually beginning any TypeScript conversion.

## How I did it

- Add `tsconfig.json` specifying recommended settings for targeting Node18, running in strict mode, permitting .js files
- Add build command to generate `.build` output
- Update Serverless settings to use TS build
- Update docker test file to build project

## How you can test it

After running `npm install --dev`, try the build with `npm run build -- --watch`. Deploying to staging and verifying that the application functions on the most basic level may be generally sufficient. 

---

Part of #187

Issue #187, opened 2024-05-29
Convert project to TypeScript


alukach:
### Core
- [x] Setup TypeScript tooling: #188
- [x] Convert Project glue code (config, db connect tooling, auth, errors, api handler): #189
- [x] Convert GraphQL TypeDefs: #195
  - Convert type-defs to GraphQL schema files
  - Setup automatic interface generation from GraphQL schema files
- [x] Convert DB Schemas: #205
- [x] Convert DB Models
  - `Batch.js` #208
  - `BatchError.js` #216
  - `Camera.js` #209
  - `Image.js` #224
  - `ImageError.ts` #211
  - `MLModel.js` #219
  - `Project.js`#220
  - `Task.js` #223
  - `User.js` #207
  - `colors.js` #210
  - `utils.js` #218
- [x] Convert GraphQL Resolvers: #229
- [x] Convert ML Handler: #232 
- [x] Convert Tasks: #233

PR #189, opened 2024-05-29
TypeScript: 2. Convert project glue code

alukach:
## What I'm changing

This PR converts the simplest files within our application to TS.  

## How I did it

- Renaming files from `.js` -> `.ts`
- Building out an interface for a `User` and `Context` are probably the most interesting parts of this PR.
- Throw an `AuthenticationError` if user is `null` when building `Context`, which is a bit of new logic but is exactly as what the Authed Model classes do shortly afterwards within the `Context` build code.
- Update tests to run against built project. This requires that the project is built before running tests (obviously), which seems non-ideal but I'm not sure of a better way (I'm a tape novice) 

## How you can test it

Deploying to staging and verifying that the application functions on the most basic level may be generally sufficient. Almost no logic code has changed.

---

Relates to #187 

> [!NOTE]
> This PR builds off of the work done in #188. As such, #188 should be merged before this PR and this PR's target should be update to `main` before merging.


jue-henry (approved):
Looks good from my walkthrough

src/api/db/connect.ts
```diff
@@ -1,34 +1,32 @@
...
 import { InternalServerError } from '../errors.js';
+import { Config } from '../../config/config.js';
 
 // TODO: consider using multiple connections (one per model)
 // to reduce risk of slow trains
 // https://mongoosejs.com/docs/connections.html#multiple_connections
 // https://thecodebarbarian.com/slow-trains-in-mongodb-and-nodejs
 
-let cachedConnectionPromise = null;
+let client: typeof mongoose | null = null;
```
nathanielrindlaub, line 10: 2 bits of confusion on this line:

1. is it ok practice to just use `typeof <some_imported_node_module`? I would have guessed that you have to explicitly import and use the Mongoose types.
2. I think this `mongoose.connect()` can also return a Promise

nathanielrindlaub: but I suppose if we're awaiting the promise it would have already resolved by the time it gets assigned to `client`... maybe answered my own question here?

src/api/db/connect.ts
```diff
@@ -1,34 +1,32 @@
...
       bufferCommands: false,
       // TODO: double check auto indexing is off in prod environment
-      autoIndex: process.env.STAGE !== 'prod'
+      autoIndex: process.env.STAGE !== 'prod',
     });
   }
 
   try {
-    const client = await cachedConnectionPromise;
     return client;
```
nathanielrindlaub, line 26: @alukach are you sure don't we still have to await this?

src/api/db/connect.ts
```diff
@@ -1,34 +1,32 @@
...
   } catch (err) {
-    throw new InternalServerError(err instanceof Error ? err : String(err));
+    throw new InternalServerError(err instanceof Error ? err.message : String(err));
```
nathanielrindlaub, line 28: @alukach I'm curious about the reason for this change?

src/api/db/connect.ts
```diff
@@ -1,34 +1,32 @@
...
 // to reduce risk of slow trains
 // https://mongoosejs.com/docs/connections.html#multiple_connections
 // https://thecodebarbarian.com/slow-trains-in-mongodb-and-nodejs
 
-let cachedConnectionPromise = null;
+let client: typeof mongoose | null = null;
 
-async function connectToDatabase(config) {
-  if (!cachedConnectionPromise) {
+async function connectToDatabase(config: Config) {
```
nathanielrindlaub, line 12: any reason we're not specify the return type here?

src/api/handler.ts
```diff
@@ -60,25 +65,51 @@ const context = async ({ event, context }) => {
...
 
 const apolloserver = new ApolloServer({
-  includeStacktraceInErrorResponses: process.env.STAGE === 'dev',
+  includeStacktraceInErrorResponses: OFFLINE_MODE || process.env.STAGE === 'dev',
   status400ForVariableCoercionErrors: true,
   typeDefs,
   resolvers,
-  csrfPrevention: true,
-  cache: 'bounded'
+  csrfPrevention: false,
```
nathanielrindlaub, line 78: hmm curious why we're turning this off.

src/api/auth/authorization.ts
```diff
@@ -1,8 +1,12 @@
 import jwt from 'jwt-simple';
+import { Config } from '../../config/config.js';
 
 const BEARER_TOKEN_PATTERN = /^Bearer [-_=.0-9a-zA-Z]+$/i;
 
-async function getUserInfo(req, config) {
+async function getUserInfo(
+  req: AWSLambda.APIGatewayProxyEvent,
```
nathanielrindlaub, line 7: From slack convo: it doesn't look like we ever import an `AWSLambda` type into this file. How does it know what that type is?

alukach: Yeah, it's a fair question. After doing a little bit of research, we see that we have the following dependency:

https://github.com/tnc-ca-geo/animl-api/blob/2e973b481f8b5f97378d83940f4d2025dac04088/package.json#L31

That dependency has the following dependency:

https://github.com/apollo-server-integrations/apollo-server-integration-aws-lambda/blob/c5a2c2dbae47b6e0b3852cea82df7799fda51529/package.json#L34

That adds the `index.d.ts` file which adds types to the [global scope](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html) (_I think..._). 
 https://github.com/DefinitelyTyped/DefinitelyTyped/blob/22683376440b9f13f263630615ca32e442b610e6/types/aws-lambda/index.d.ts

Anyways, I'll update this to an import to be a bit more explicit.

src/api/db/connect.ts
alukach: According to the TS types, `mongoos.connect()` returns a promise of a client. Once we await that, there's no reason to await it again.  However, we never hit a bug because you can await non-promises as well:

```js
> async function foo() {
... const value = await 99;
... return value;
... }
undefined
> foo().then(console.log)
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 431,
  [Symbol(trigger_async_id_symbol)]: 428
}
> 99
```

src/api/db/connect.ts
alukach: The constructor of our `InternalServerError` expects a `string` argument, not an `Error` instance:

https://github.com/tnc-ca-geo/animl-api/blob/2e973b481f8b5f97378d83940f4d2025dac04088/src/api/errors.js#L8-L9

src/api/handler.ts
alukach: There was some cruft that I mistakenly left in this PR from trying to get local development to work. Removed in [d2efca9](https://github.com/tnc-ca-geo/animl-api/pull/189/commits/d2efca993fd1107776272d06ff5c643435878bcb)

src/api/db/connect.ts
alukach: Not really.  I think I just figured it wasn't really necessary in that we don't need it when we await it here:

https://github.com/tnc-ca-geo/animl-api/blob/d2efca993fd1107776272d06ff5c643435878bcb/src/api/handler.ts#L47

However, I added an export to be more explicit: 4c6a5f792f75ba4cda5744c0f731bf5de1e587e8

src/api/db/connect.ts
nathanielrindlaub: ah, ok makes sense!

src/api/auth/authorization.ts
nathanielrindlaub: Got it, ok sweet thanks @alukach

tsconfig.json
```diff
@@ -6,6 +6,10 @@
     "sourceMap": true,
     "allowJs": true,
     "outDir": "./.build",
+    
+    // Use CJS for better interoperability with legacy modules (e.g. @as-integrations/aws-lambda)
+    "module": "CommonJS",
+    "moduleResolution": "node",
```
nathanielrindlaub, line 12: Hey @alukach, the tests are failing now as the commit this change was a part of, and at a glance this line seems like the most likely culprit. I don't know enough about issue with legacy modules to feel comfortable messing with it, so just wanted to flag it for you.

nathanielrindlaub: It seems like maybe `"moduleResolution": "node16"` or `"nodenext" should support but ES2020 or CommonJS modules? https://www.typescriptlang.org/tsconfig/#node16nodenext

nathanielrindlaub: Perhaps related: I just deployed to staging and got the following error when I tried to open the staging frontend:

```
{
    "errorType": "ReferenceError",
    "errorMessage": "exports is not defined in ES module scope\nThis file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains \"type\": \"module\". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.",
    "stack": [
        "ReferenceError: exports is not defined in ES module scope",
        "This file is being treated as an ES module because it has a '.js' file extension and '/var/task/package.json' contains \"type\": \"module\". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.",
        "    at file:///var/task/src/api/handler.js:5:23",
        "    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)",
        "    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)",
        "    at async /opt/sls-sdk-node/wrapper.js:13440:14",
        "    at async _loadUserApp (file:///var/runtime/index.mjs:1081:16)",
        "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1119:21)",
        "    at async start (file:///var/runtime/index.mjs:1282:23)",
        "    at async file:///var/runtime/index.mjs:1288:1"
    ]
}
```

nathanielrindlaub (changes requested):
@alukach looking really close, but I think something related to the changes to the `tsconfig.json` broke this build.

PR #195, opened 2024-06-03
TypeScript: 3. Auto-generate TS interfaces from GraphQL type-defs

alukach:
## What I'm changing

This PR converts our GraphQL types to TS files and configures tooling to auto-generate TypeScript interfaces for the GraphQL API.

## How I did it

This looks very easy but it took me quite a while to get right.  It's one of those things that can be done correctly one way and done incorrectly many many ways. 🤕 

1. Convert all `type-defs/*` files to `.ts`
2. Add ` /* GraphQL */ ` to the GraphQL type definitions. This instructs modern code editors (eg VSCode) and Prettier to style the strings as GraphQL documents. More importantly, it also informs `@graphql-codegen` that the strings are GraphQL documents. This seems like a cosmetic detail, but is critical in order for the GraphQL Codegen to work properly.
3. Add `@graphql-codegen` dependency and `codegen.ts` script in order to auto-generate types in `src/@types/graphql.ts`. Setup `npm run codegen` to run script.
4. Add Github actions workflow to verify that generated GraphQL types are committed to the repo

## How you can test it

1. `npm install --dev`
2. `npm run codegen`

---

relates to #187

> [!NOTE]
> This PR builds off of the work done in #189. As such, #189 should be merged before this PR and this PR's target should be update to `main` before merging.


src/@types/graphql.ts
```diff
@@ -0,0 +1,1192 @@
...
+  Boolean: { input: boolean; output: boolean; }
+  Int: { input: number; output: number; }
+  Float: { input: number; output: number; }
+  Date: { input: any; output: any; }
+  JSONObject: { input: any; output: any; }
+};
+
+export type AutomationAction = {
+  __typename?: 'AutomationAction';
+  alertRecipients?: Maybe<Array<Maybe<Scalars['String']['output']>>>;
```
jue-henry, line 21: If these fields are nullable, then why do we need this `Maybe` type?

src/@types/graphql.ts
```diff
@@ -0,0 +1,1192 @@
+export type Maybe<T> = T | null;
```
jue-henry, line 1: Just for my own understanding, we are basically defining types for all our graphql requests and responses? Was this auto-generated?

jue-henry: So far these types aren't actually being used anywhere right? This PR just setups the generation of these types right?

alukach: Currently, they're not. However, they will be used to describe expected input/output data in our resolvers (see #196)

src/@types/graphql.ts
alukach: That's a TS representation of how it's currently typed here: https://github.com/tnc-ca-geo/animl-api/blob/cb48e5dc149d01643b60c038837204d256180cce/src/api/type-defs/objects/AutomationRule.ts#L7-L13

This means that all of these are valid `AutomationAction` payloads:


```json5
{
  // ...
  alertRecipients: ["someRecipient@mail.com"]
}
```

```json5
{
  // ...
  alertRecipients: undefined
}
```


```json5
{
  // ...
  alertRecipients: []
}
```

```json5
{
  // ...
  alertRecipients: null
}
```

```json5
{
  // ...
  alertRecipients: [null]
}
```

Personally, I doubt that some of these are actually valid (particularly the last one).

The type should probably be specified as:

```gql
   type AutomationAction { 
     type: String! 
     alertRecipients: [String!]! 
     mlModel: String 
     confThreshold: Float 
     categoryConfig: JSONObject 
   } 
```

However I'm trying to avoid critiquing types unless they cause functional issues within the codebase's type completeness.

alukach: In #197, #198, and #199 I had to update the types to make TS happy.

jue-henry: Gotcha

nathanielrindlaub: @alukach I think your policy of trying to avoid modifying the graphql types unless they cause issues is a good one for now. I have no doubt the graphql types can and should be tightened up but it could easily turn into a can of worms that produces some unexpected breaking issues. 

For example, in this case, an `AutomationAction` is used to provide details on how to execute an [Automation Rule](https://docs.animl.camera/fundamentals/automation-rules) for both `type: "run-inference"` and `type: "send-alert"` actions, and some of those props are only relevant to sending alerts (`alertRecipients`) while some are only relevant to requesting inference (`mlModel`, `confThreshold`, `categoryConfig`). So `alertRecipients` should not be required for inference-request-related `AutomationActions`, but perhaps if it's present the array should always have stings? So the type-def should be something like `alertRecipients: [String!]`?

For now, don't worry about this though. Not requesting a change just adding some context.

PR #196, opened 2024-06-03
[IN PROGRESS] TypeScript: 4. Convert GraphQL resolvers to TS

alukach:
## What I'm changing

This PR continues the GraphQL conversion towards TS, converting the resolvers. These types makes use of the auto-generated types introduced in #195.

## How I did it

This PR is a mostly straight-forward conversion.  However, `src/api/resolvers/Scalars.ts` required new strict type-checking before marshaling data between types. 

Additionally, I was required to change `import GraphQLJSON from 'graphql-type-json'` to `import * as GraphQLJSON from 'graphql-type-json'` in order to conform to the types reported by the `graphql-type-json` module. This was a bit concerning and should be tested.

## How you can test it

Deploy to staging, test API endpoints that work with a DateTime field.

---

part of #188

> [!NOTE]
> This PR builds off of the work done in #195. As such, #195 should be merged before this PR and this PR's target should be update to `main` before merging.


alukach:
I'm going to close this to do a bit more work with the resolvers, adding return types.

src/api/resolvers/Scalars.ts
```diff
@@ -1,6 +1,6 @@
 import { DateTime } from 'luxon';
 import { GraphQLScalarType } from 'graphql';
-import GraphQLJSON from 'graphql-type-json';
+import * as GraphQLJSON from 'graphql-type-json';
```
alukach, line 3: I think this is functionally equivalent but worth confirming. Needed to do this to make TS happy.

src/api/resolvers/Scalars.ts
```diff
@@ -12,22 +12,28 @@ const Date = new GraphQLScalarType({
...
     // e.g. input is passed into query as a JSON variable
     return DateTime.fromISO(value);
   },
   serialize(value) {
+    if (!isDate(value)) throw new Error(`Expected string, got ${typeof value}`);
     // Prep return value to be sent to client
     return DateTime.fromJSDate(value).toISO();
   },
   parseLiteral(ast) {
+    if (ast.kind !== 'StringValue') throw new Error(`Expected string, got ${ast.kind}`);
```
alukach, line 26: Needed to up validation to make TS happy.

PR #197, opened 2024-06-04
TypeScript: Resolvers(query.imageErrors): Add count to ImageErrors pageInfo object

alukach:
Currently, in GraphQL, our types have us returning an `ImageErrorsConnection` from `Query.ImageErrors`:
https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/type-defs/root/Query.js#L10

That type has a `PageInfo` object:
https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/type-defs/payloads/ImageErrorsConnection.js#L2-L3

Which does not contain a `count` property:
https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/type-defs/objects/PageInfo.js#L2-L7

However, in our resolver, we _are_ returning a `count`:
https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/resolvers/Query.js#L68-L78

This PR adds a `count` to the response type to appease TypeScript.

### DIFF
--- a/src/api/type-defs/objects/PageInfo.ts
+++ b/src/api/type-defs/objects/PageInfo.ts
@@ -6,11 +6,13 @@ export default /* GraphQL */ `
     hasNext: Boolean
   }
 
-  type PageCount {
+  type PageInfoWithCount implements PageInfo {
+    previous: String
+    hasPrevious: Boolean
+    next: String
+    hasNext: Boolean
     count: Int
   }
-
-  union PageInfoWithCount = PageInfo | PageCount
 `;
 
 // TODO: should any of these be required?
--- a/src/api/type-defs/objects/PageInfo.ts
+++ b/src/api/type-defs/objects/PageInfo.ts
@@ -1,12 +1,19 @@
 export default /* GraphQL */ `
-  type PageInfo {
+  interface IPageInfo {
+    previous: String
+    hasPrevious: Boolean
+    next: String
+    hasNext: Boolean
+  }
+  
+  type PageInfo implements IPageInfo {
     previous: String
     hasPrevious: Boolean
     next: String
     hasNext: Boolean
   }
 
-  type PageInfoWithCount implements PageInfo {
+  type PageInfoWithCount implements IPageInfo {
     previous: String
     hasPrevious: Boolean
     next: String
--- a/src/api/type-defs/objects/PageInfo.ts
+++ b/src/api/type-defs/objects/PageInfo.ts
@@ -5,6 +5,12 @@ export default /* GraphQL */ `
     next: String
     hasNext: Boolean
   }
+
+  type PageCount {
+    count: Int
+  }
+
+  union PageInfoWithCount = PageInfo | PageCount
 `;
 
 // TODO: should any of these be required?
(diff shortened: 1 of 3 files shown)


src/api/type-defs/objects/PageInfo.ts
```diff
@@ -5,6 +5,12 @@ export default /* GraphQL */ `
     next: String
     hasNext: Boolean
   }
+
+  type PageCount {
+    count: Int
+  }
+
+  union PageInfoWithCount = PageInfo | PageCount
```
nathanielrindlaub, line 13: @alukach I may be misunderstanding GraphQL unions, but my understanding is that the `union PageInfoWithCount = PageInfo | PageCount` means that it can be either `PageInfo` or `PageCount` types, but I think what we want is a way to extend the base `PageInfo` properties to have an additional `count` property for use in `ImageErrorsConnection`. 
 
If you check out the [return payload](https://github.com/tnc-ca-geo/animl-api/blob/eead714ee93d7cb2b5e6372e72dd024f3af24c9c/src/api/resolvers/Query.js#L78) from the `imageErrors` resolver, it includes all of the `PageInfo` fields plus an additional `count` field.

nathanielrindlaub (commented):
Hey @alukach I have a question here - could be just my lack of familiarity with unions but I don't that that's what we want?

src/api/type-defs/objects/PageInfo.ts
nathanielrindlaub: Also it seems to be failing the tests.

alukach: Good catch, union doesn't mean extends, my mistake. I've updated this in [9caf1e5](https://github.com/tnc-ca-geo/animl-api/pull/197/commits/9caf1e5c5091f681b487fbc2d99eb3d242f61fe5)

nathanielrindlaub: I just did some more reading and I have to say I don't care for this one bit, but apparently GraphQL [doesn't support the ability to simply extend a type](https://stackoverflow.com/questions/48917863/how-do-you-extend-types-in-graphql) and call it something else. The GraphQL concept of `interfaces` is also kind of lame IMO, but I think we need to tweak what you have to look like this to properly use and `implement` an interface:

```javascript
export default /* GraphQL */ `
  // the shared properties are the "interface"
  interface Pagination {
    previous: String
    hasPrevious: Boolean
    next: String
    hasNext: Boolean
  }

  // still need to write out all the shared interface props again for some reason
  type PageInfo implements Pagination {
    previous: String
    hasPrevious: Boolean
    next: String
    hasNext: Boolean
  }

  // this one looks good
  type PageInfoWithCount implements Pagination {
    previous: String
    hasPrevious: Boolean
    next: String
    hasNext: Boolean
    count: Int
  }
`;
```

I don't really understand what the point of `interfaces` are other than to yell at you if you add a new prop to the interface but forget to add it to the actual types that implement them. But alas. Here are the docs I was referencing: https://graphql.com/learn/interfaces-and-unions/

PR #198, opened 2024-06-04
TypeScript: Resolvers(query.mlModels): Mark MLModel.categories as an array of non-nullable values

alukach:
In our codebase, we have the MLModel's `categories` property as an array of either `Categories` or `null`:

https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/type-defs/objects/MLModel.js#L8-L14

However, in our model, we show `Categories` as being only of a `Category`:

https://github.com/tnc-ca-geo/animl-api/blob/5522f6a70c3cc9d0eddf7a5f75b673626f53bc27/src/api/db/schemas/MLModel.js#L25-L37

This is causing the following TS error in following codeE:

https://github.com/tnc-ca-geo/animl-api/blob/286f7e693bd4078c709754ade5c667adbc923643/src/api/resolvers/Query.ts#L131-L137

```
Type '(Document<unknown, {}, { _id: string; version: string; defaultConfThreshold: number; categories: DocumentArray<{ _id: string; name: string; color: string; }>; description?: string | null | undefined; }> & { ...; } & Required<...>)[]' is not assignable to type 'MlModel[]'.
  Type 'Document<unknown, {}, { _id: string; version: string; defaultConfThreshold: number; categories: DocumentArray<{ _id: string; name: string; color: string; }>; description?: string | null | undefined; }> & { ...; } & Required<...>' is not assignable to type 'MlModel'.
    The types of 'categories.push' are incompatible between these types.
      Type '(...args: (AnyKeys<{ _id: string; name: string; color: string; }> & AnyObject)[]) => number' is not assignable to type '(...items: Maybe<Categories>[]) => number'.
        Types of parameters 'args' and 'items' are incompatible.
          Type 'Maybe<Categories>' is not assignable to type 'AnyKeys<{ _id: string; name: string; color: string; }> & AnyObject'.
            Type 'null' is not assignable to type 'AnyKeys<{ _id: string; name: string; color: string; }> & AnyObject'.
              Type 'null' is not assignable to type 'AnyKeys<{ _id: string; name: string; color: string; }>'.ts(2322)
```

This PR fixes that.


src/api/type-defs/objects/MLModel.ts
```diff
@@ -10,6 +10,6 @@ export default /* GraphQL */ `
     description: String
     version: String!
     defaultConfThreshold: Float
-    categories: [Categories]
+    categories: [Categories!]
```
nathanielrindlaub, line 13: @alukach I think the plural thing was a typo:

```
  type Categories { // this should be Catagory
    _id: String!
    name: String!
    color: String!
  }
```

and the `type MLModel` should be (`categories` should also be required):
```
categories: [Category!]!
```

nathanielrindlaub (changes requested):

nathanielrindlaub:
Yeah go for it!

src/api/type-defs/objects/MLModel.ts
nathanielrindlaub: Yes please! Thanks @alukach

alukach: @nathanielrindlaub Okay, done ✅

nathanielrindlaub: Mind adding one more explaination mark to make `categories` required? I know that wasn't the case before but MLModels should always have categories (see note above). 

```
categories: [Category!]!
```

alukach: Sounds good @nathanielrindlaub , see https://github.com/tnc-ca-geo/animl-api/pull/198/commits/3ff48e228ad0cd63fa28fa756d4295b1bf5a5e0b

PR #199, opened 2024-06-04
TypeScript: Resolvers(query.tasks): Add PageInfo for Query.tasks response type

alukach:
Current code returns a `pageInfo` property in `Query.tasks`, but we don't include that in the `TasksPayload` specified in the types:

https://github.com/tnc-ca-geo/animl-api/blob/27b3e3d955208ffdaa061aeced28689e76644ad4/src/api/type-defs/root/Query.ts#L4

https://github.com/tnc-ca-geo/animl-api/blob/286f7e693bd4078c709754ade5c667adbc923643/src/api/resolvers/Query.ts#L21-L37

### DIFF
--- a/src/@types/graphql.ts
+++ b/src/@types/graphql.ts
@@ -1047,6 +1047,7 @@ export type Task = {
 
 export type TasksPayload = {
   __typename?: 'TasksPayload';
+  pageInfo: PageInfo;
   tasks: Array<Maybe<Task>>;
 };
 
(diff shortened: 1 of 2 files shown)


nathanielrindlaub (changes requested):
This looks good to me @alukach, we just need to get those tests passing and I can merge it.

alukach:
@nathanielrindlaub There are a lot of independent reasons that the builds are failing, this is just a fix of one of those.  All of these `TypeScript: Resolvers(*)...` PRs represent a single one of those fixes. The target branch of this PR is [feature/typescript/graphql-resolvers](https://github.com/tnc-ca-geo/animl-api/tree/feature/typescript/graphql-resolvers), so we can merge them all into that branch and then when we finally get the tests/build success, we can merge #196 into `main`.

PR #200, opened 2024-06-05
TypeScript: Resolvers(query.task/stats/exportErrors/exportAnnotations): Return toObject() in order to convert `_id` to string

alukach:
For any method where a `Task` is returned, we see the following error:

```ts
Type 'Document<unknown, {}, { type: "GetStats" | "ExportAnnotations" | "ExportImageErrors" | "CreateDeployment" | "UpdateDeployment" | "DeleteDeployment"; user: string; projectId: string; status: "SUBMITTED" | ... 2 more ... | "COMPLETE"; created: Date; updated: Date; output?: any; }> & { ...; } & { ...; }' is not assignable to type 'Task'.
  Types of property '_id' are incompatible.
    Type 'ObjectId' is not assignable to type 'string'.ts(2322)
```

This seems to be an issue where GraphQL expected the `_id` to be a string:

https://github.com/tnc-ca-geo/animl-api/blob/27b3e3d955208ffdaa061aeced28689e76644ad4/src/@types/graphql.ts#L1038
https://github.com/tnc-ca-geo/animl-api/blob/27b3e3d955208ffdaa061aeced28689e76644ad4/src/@types/graphql.ts#L10

However, Mongoose shows it as an `mongoose.Types.ObjectId`. Calling `.toObject()` on the response object rectifies this.  This may become a common pattern for our resolvers...  I would like to imagine that there's a better way to handle this, but I'm not sure what that would be, perhaps something with the Scalars for the output?

### DIFF
--- a/src/api/resolvers/Query.ts
+++ b/src/api/resolvers/Query.ts
@@ -7,7 +7,8 @@ export default {
     { input }: gql.QueryProjectsArgs,
     context: Context,
   ): Promise<gql.Project[]> => {
-    return context.models.Project.getProjects(input, context);
+    const projects = await context.models.Project.getProjects(input, context);
+    return projects.map((project) => project.toObject());
   },
 
   users: async (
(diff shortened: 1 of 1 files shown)


alukach:
I actually think the reason that this conversion works is because of https://github.com/Automattic/mongoose/issues/12883, which may actually be a bad solution...

PR #202, opened 2024-06-05
TypeScript: Resolvers: Update codegen to use mongoose.Types.ObjectID for ID scalar 

alukach:


src/api/db/schemas/Task.ts
```diff
@@ -0,0 +1,53 @@
...
+  output: { type: Object },
+});
+
+TaskSchema.plugin(MongoPaging.mongoosePlugin);
+
+export default mongoose.model<Task>('Task', TaskSchema);
+
+// The `type` property of the TaskSchema seems to be causing problems with the automatic
+// type generation, so we manually create our own type.
+interface Task {
```
nathanielrindlaub, line 38: I would prefer not to manually type this this for consistency and maintainability reasons. It seems crazy that the Mongoose type generation can't handle a property named `type`. We have `type` props in other places too: https://github.com/tnc-ca-geo/animl-api/blob/f229e6ba13da0ffc47aa91cf00ceb84d42aaaf2c/src/api/db/schemas/Project.js#L10

src/@types/mongo-cursor-pagination.d.ts
```diff
@@ -0,0 +1,24 @@
+// https://github.com/mixmaxhq/mongo-cursor-pagination
+declare module 'mongo-cursor-pagination' {
```
nathanielrindlaub, line 2: Just out of curiosity, what is the `d` in the `.d.ts` file extension all about?

nathanielrindlaub (changes requested):
I'm a little confused on this one b/c the name of the PR doesn't seem to match the code changes, but I would prefer to find a different solution to manually typing some Mongoose Schemas and auto-generating others. I feel like it will be hard to remember and maintain.

alukach:
@nathanielrindlaub sorry for the confusion, I created this PR but left it in the `draft` state without reviewers as it was still a bit of a work in progress.

I agree that we don't want to create custom types for our interfaces. However, I have been hitting strange type errors with both the `Task` and `Project` schemas (both contain properties named `type`). I am honestly unsure if it is the `type` property that is the cause of the errors (or at least related), or if this is just a coincidence. In this PR, I've been able to get around the `Task` errors by letting GraphQL->TS codegen know that an `_id` can be either a `string` or `ObjectId`.

TS is mature enough that tools like Mongoose have good support for inferring types based on inputted schemas, however this isn't always perfect. Providing custom types to give it a hand is something we may need to reach for, but hopefully not.

With regards to the `.d.ts` file, this is a way to specify a global type file within TypeScript, so it doesn't need to be imported explicitely. This allows us to specify types for systems that we don't actually control. Pretty much the same this as in [this convo](https://github.com/tnc-ca-geo/animl-api/pull/189#discussion_r1626660032), however in this case we're adding the types rather than `@types/aws-lambda` adding them.

alukach:
@nathanielrindlaub I'm going to remove the `mongodb-cursor-pagination` types from this PR as we don't need them just yet, but that change will be coming down the pipe (once we start converting the schema files).

nathanielrindlaub (approved):
Looks good to me @alukach! Thanks for figuring this out and explaining it so clearly!

PR #206, opened 2024-06-11
[Feature PR] TypeScript: 4. Convert GraphQL resolvers to TS

alukach:
> [!IMPORTANT]
> This PR consists of commits only from PRs approved elsewhere. Once CICD passes, it should be considered approved.

## What I'm changing

This PR continues the GraphQL conversion towards TS, converting the resolvers. These types makes use of the auto-generated types introduced in #195.

## How I did it

This PR is a mostly straight-forward conversion.  However, `src/api/resolvers/Scalars.ts` required new strict type-checking before marshaling data between types. 

Additionally, I was required to change `import GraphQLJSON from 'graphql-type-json'` to `import * as GraphQLJSON from 'graphql-type-json'` in order to conform to the types reported by the `graphql-type-json` module. This was a bit concerning and should be tested.

Additional contributions made through separate PRs:

- #197
- #198
- #199
- #201
- #202

## How you can test it

Deploy to staging, test API endpoints that work with a DateTime field.

---

part of #188


src/api/resolvers/Scalars.ts
```diff
@@ -1,6 +1,6 @@
 import { DateTime } from 'luxon';
 import { GraphQLScalarType } from 'graphql';
-import GraphQLJSON from 'graphql-type-json';
+import * as GraphQLJSON from 'graphql-type-json';
```
alukach, line 3: I believe that this change has no impact, however this should be verified.

alukach:
This PR will be made ready via #229.

PR #207, opened 2024-06-12
TypeScript: Models(User)

alukach:
## What I'm changing

This PR converts the User model to TS

## How I did it

1. Rename file from `.js` to `.ts`
2. Create `util-models.ts` with some model utilities. Namely:
    1. `BaseAuthedModel` is to be used by all the authenticated models to reduce boilerplate code.
    2. `roleCheck` decorator to make it simpler to verify user permissions. Using this decorator requires us to enable `experimentalDecorators` in the tsconfig.
    3. `MethodParams` utility type to easily extract arguments from class methods. Useful to minimize writing duplicate types when writing pass-through methods from an authed model to a non-authed model.

## How you can test it

Deploy to staging, test that:
* `Query.users` query works when authenticated
* `Query.users` query throws error when not authenticated

---

Part of #187


src/api/db/models/User.ts
```diff
@@ -0,0 +1,277 @@
...
+  updated: Date | undefined;
+  enabled: boolean | undefined;
+  status: Cognito.UserStatusType | undefined;
+}
+
+interface CreateGroupsInput {
+  name: string;
+}
+
+interface CreateUserInput {
```
nathanielrindlaub, line 267: @alukach aren't some of these input types already being generated by the GraphQL codegen?

alukach: Yeah, pretty much the same thing as https://github.com/tnc-ca-geo/animl-api/pull/208#discussion_r1640172109.

In [ecf7fb1](https://github.com/tnc-ca-geo/animl-api/pull/207/commits/ecf7fb1905aef16a3062ead26b353987ca7a9a8e), I've reworked this to utilize the generated GQL types.

nathanielrindlaub (approved):
Awesome this looks great @alukach. Merging.

PR #209, opened 2024-06-12
TypeScript: Models(Camera)

alukach:
## What I'm changing

This PR converts the Camera model to TS.

## How I did it

There's a pattern throughout these classes that I have been removing. In this pattern, there is an operation function that is run with `retry`, like so:

```ts
const operation = async (input) => {
  return await retry(async (bail, attempt) => {
    // ... do something
  }, { retries: 2 });
};

try {
  await operation(someInput);
  // ...
  return { isOk: true };
} catch (err) {
  if (err instanceof GraphQLError) throw err;
  throw new InternalServerError(err);
}
```

This makes it a bit more onerous to write types for, as we need to type the input for the `operation` function.  A simpler pattern looks like the following, where we replace the `operation` call with the contents of the function:

```ts
try {
  await retry(async (bail, attempt) => {
    // ... do something
  }, { retries: 2 });
  // ...
  return { isOk: true };
} catch (err) {
  if (err instanceof GraphQLError) throw err;
  throw new InternalServerError(err);
}
```

#### Before
```ts
    const saveWirelessCamera = async (input) => {
      const { projectId, cameraId, make, model } = input;
      return await retry(async () => {
        const newCamera = new WirelessCamera({
          _id: cameraId,
          make,
          projRegistrations: [{ projectId, active: true }],
          ...(model && { model })
        });
        await newCamera.save();
        return newCamera;
      }, { retries: 2 });
    };

    try {
      // create Wireless Camera record
      const camera = await saveWirelessCamera({ ...input, projectId });
```
#### After
```ts
    try {
      // create Wireless Camera record
      const camera = await await retry(
        async () => {
          const newCamera = new WirelessCamera({
            _id: input.cameraId,
            make: input.make,
            projRegistrations: [{ projectId, active: true }],
            model: input.model,
          });
          await newCamera.save();
          return newCamera;
        },
        { retries: 2 },
      );
```

---

Part of #187


src/api/db/models/Camera.ts
```diff
@@ -104,44 +123,45 @@ export class CameraModel {
...
-      throw new InternalServerError(err);
+      throw new InternalServerError(err as string);
     }
   }
 
-  static async unregisterCamera(input, context) {
-    const successfulOps = [];
+  static async unregisterCamera(
+    input: { cameraId: string },
+    context: Context,
```
nathanielrindlaub, line 159: Not that it matters a ton,  but this one is autogenerated and exported from `src/@types/graphql.ts` as `UnregisterCameraInput`.

nathanielrindlaub (approved):

PR #216, opened 2024-06-13
TypeScript: Models(BatchError)

alukach:
Convert BatchError to TS.  A pretty straight-forward conversion.

---

Part of #187


src/api/db/models/BatchError.ts
```diff
@@ -0,0 +1,73 @@
...
+  static async createError(input: { batch: string; error: string }): Promise<BatchErrorSchema> {
+    try {
+      return await retry(
+        async () => {
+          const newBatchError = new BatchError({
+            batch: input.batch,
+            error: input.error,
+          });
+          await newBatchError.save();
+          return newBatchError;
```
jue-henry, line 29: Just pointing out that we're returning the entire `batchError` instead of the specific fields listed previously. I assume its ok but in case there are fields we don't want to return. However I do like the rewrite

jue-henry (approved):

nathanielrindlaub (approved):

PR #172, opened 2024-04-01
Adding reviewed field

jue-henry:
Addressing issue #161, we created a new field on the images collection, in order to make it easier to search and filter for. In order to do this, we also added a method to be ran after relevant db operations are done to keep this field up to date.

PR #218, opened 2024-06-14
[IN PROGRESS] TypeScript: Models(utils)

alukach:


src/api/db/models/utils.ts
```diff
@@ -0,0 +1,705 @@
...
+        dateAdded: {
+          ...(addedStart && { $gte: addedStart.toJSDate() }),
+          ...(addedEnd && { $lt: addedEnd.toJSDate() }),
+        },
+      },
+    });
+  }
+
+  // match notReivewed filter
+  if (reviewed === false) {
```
jue-henry, line 186: Looks like this code is out of sync with models/utils.js. Please see: https://github.com/tnc-ca-geo/animl-api/blob/03b284db07a2e2753f9e122f6bcf8c7d0bf82f5a/src/api/db/models/utils.js#L147

jue-henry (changes requested):

src/api/db/models/utils.ts
jue-henry: @alukach Also I performed a merge with main instead of a rebase just to resolve a merge conflict with main instead of rebasing just because I was less familiar with the changes made. Feel free to revert the merge and do a proper rebase

alukach: Good catch.  That's a bit concerning, my `utils.ts` must have been created from `utils.js` before #172 was merged to `main`?  Not sure how that would have happened...

alukach: Should be fixed in [6d375a0](https://github.com/tnc-ca-geo/animl-api/pull/218/commits/6d375a07a010a444cf0914379fb68317c89124b3)

nathanielrindlaub: @jue-henry can you confirm that @alukach pulled in the latest `utils.js` code into the new `utils.ts`?

jue-henry: yup looks good now

nathanielrindlaub (approved):

PR #219, opened 2024-06-14
TypeScript: Models(MLModel)

alukach:


src/api/db/models/MLModel.ts
```diff
@@ -0,0 +1,60 @@
+import GraphQLError, { InternalServerError, NotFoundError } from '../../errors.js';
+import MLModel, { MLModelSchema } from '../schemas/MLModel.js';
+import retry from 'async-retry';
+import { BaseAuthedModel, MethodParams } from './utils-model.js';
+import { HydratedDocument } from 'mongoose';
+
+export class MLModelModel {
+  static async queryById(_id: string): Promise<HydratedDocument<MLModelSchema>> {
+    const query = { _id };
```
nathanielrindlaub, line 9: Not sure why [CodeQL complained about this](https://github.com/tnc-ca-geo/animl-api/pull/220#discussion_r1642957203) in the `Project` model and not here but might as well preempt it with the same fix?

nathanielrindlaub (changes requested):
I think just a one liner update and we can merge this one @alukach.

PR #220, opened 2024-06-14
TypeScript: Models(Project)

alukach:
More of the usual.

---

Part of #187


src/api/db/models/Project.ts
```diff
@@ -0,0 +1,746 @@
...
+        { retries: 2 },
+      );
+    } catch (err) {
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  static async createView(
+    input: ViewSchema,
```
nathanielrindlaub, line 230: I think `gql.CreateViewInput` would be more appropriate here, although the properties are identical at the moment. But in the future there's always the potential that they diverge.

src/api/db/models/Project.ts
```diff
@@ -0,0 +1,746 @@
...
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  // NOTE: this function is called by the task handler
+  static async createDeployment(
+    input: gql.CreateDeploymentInput,
+    context: Context,
+  ): Promise<HydratedCameraConfig> {
```
nathanielrindlaub, line 430: Correct me if I'm wrong here, but because `CameraConfigs` are subdocuments of `Projects`, shouldn't we return `Promise<HydratedSingleSubdocument<CameraConfigSchema>>>`, in the same way we do in `CreateView()` above?

src/api/db/models/Project.ts
```diff
@@ -0,0 +1,746 @@
...
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  // NOTE: this function is called by the task handler
+  static async deleteDeployment(
+    { cameraId, deploymentId }: gql.DeleteDeploymentInput,
+    context: Context,
+  ): Promise<HydratedCameraConfig> {
```
nathanielrindlaub, line 544: It looks like we return an object with both the updated `project` and `camConfig` in this method. Also see note about using `<HydratedSingleSubdocument<CameraConfigSchema>>` above.

src/api/db/models/Project.ts
```diff
@@ -0,0 +1,746 @@
...
+    } catch (err) {
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  static async createLabel(
+    input: gql.CreateProjectLabelInput,
+    context: Context,
+  ): Promise<HydratedProject['labels'][0]> {
```
nathanielrindlaub, line 579: In this case, because we don't appear to nee to use any of Mongoose methods on this subdocument, I think it's ok to return the "Nested Path" type rather than the Hydrated Subdocument type? Perhaps that's true in other places as well.

nathanielrindlaub: No change requested just an observation.

src/api/db/models/Project.ts
```diff
@@ -0,0 +1,746 @@
...
+
+        return {
+          ...view,
+          filters: {
+            ...view.filters,
+            labels: view.filters.labels.filter((label) =>
+              project.labels.some((l) => l._id === label),
+            ),
+          },
+        } as mongoose.Types.Subdocument<mongoose.Types.ObjectId> & ViewSchema;
```
nathanielrindlaub, line 645: @alukach does this mean that if we don't explicitly add an `_id` field to a subdocument, and instead rely on Mongoose to add it for us, mongoose's `InferRawDocType` doesn't know about the `_id`? I was wondering this (question 2.a) [here](https://github.com/tnc-ca-geo/animl-api/issues/221).

nathanielrindlaub: Also are you sure we're using that `mongoose.Types.Subdocument<>` generic correctly? I'm a little confused why we'd just be passing in the `ObjectId` into it and not the entire `ViewSchema`.

nathanielrindlaub (changes requested):
Hey @alukach looking solid, but I have a few questions here. Most of them are related to the Mongoose/typescript questions I articulated here: https://github.com/tnc-ca-geo/animl-api/issues/221

src/api/db/models/Project.ts
alukach: I think I was just being a bit sloppy at that moment, properly typing the output as done in [b802430](https://github.com/tnc-ca-geo/animl-api/pull/220/commits/b8024302211ff73c8b8f073c8d36652147634ee6) is probably the more correct way for us to go.

src/api/db/models/Project.ts
alukach: To be honest, I'm not totally certain as to why this was necessary when I wrote it.  I assume I was reaching for some things to make TS happy for the next line:

https://github.com/tnc-ca-geo/animl-api/blob/b8024302211ff73c8b8f073c8d36652147634ee6/src/api/db/models/Project.ts#L653

However, removing the forced coercion, we still get the same type:

image


To answer your actual question about the generic usage, the first arg to the `mongoose.Types.Subdocument` is actually the `idType`: 

https://github.com/Automattic/mongoose/blob/17334785f52e001610200380f10416806c674910/types/types.d.ts#L86-L97

src/api/db/models/Project.ts
alukach: While we do return both a `project` and `camConfig` from the retryable portion of this method, the actual method only returns the `camConfig`:

https://github.com/tnc-ca-geo/animl-api/blob/f9797eb914a6a8662d0fb891f7536a13657fbe89/src/api/db/models/Project.ts#L563-L569

src/api/db/models/Project.ts
alukach: I think that's probably more correct, however I admit that I don't have a full grasp on what exactly the _functional_ differences are between `HydratedSingleSubdocument` and `HydratedDocument`

src/api/db/models/Project.ts
alukach: Fixed

src/api/db/models/Project.ts
nathanielrindlaub: Good call!

nathanielrindlaub (approved):

PR #222, opened 2024-06-17
Mark all values in arrays as required

alukach:
## What I'm changing

I get the impression that any array argument we take in within our GraphQL API should only contain records of a specific type, never `null` values.  This PR ensures that all arrays (optional or required) expects non-null values.

## How I did it

Find and replace.


nathanielrindlaub (approved):
Looks good to me! Approving and merging now.

PR #232, opened 2024-06-27
TypeScript: Migrate ML handlers

alukach:
Convert the ML handlers to TS.

---

Part of #187

### DIFF
--- a/src/ml/handler.ts
+++ b/src/ml/handler.ts
@@ -1,8 +1,13 @@
 import { GraphQLClient, gql } from 'graphql-request';
-import { modelInterfaces } from './modelInterfaces.js';
-import { getConfig } from '../config/config.js';
+import { type Detection, modelInterfaces } from './modelInterfaces.js';
+import { type Config, getConfig } from '../config/config.js';
+import { type GraphQLError } from 'graphql';
+import * as gqlTypes from '../@types/graphql.js';
 
-async function requestCreateInternalLabels(input, config) {
+async function requestCreateInternalLabels(
+  input: { labels: Detection[] },
+  config: Config,
+): Promise<gqlTypes.StandardPayload> {
   const variables = { input: input };
   const mutation = gql`
     mutation CreateInternalLabels($input: CreateInternalLabelsInput!) {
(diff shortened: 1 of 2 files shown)


src/ml/handler.ts
```diff
@@ -12,78 +17,81 @@ async function requestCreateInternalLabels(input, config) {
...
+    event.Records.map((record) => {
+      return singleInference(config, record);
+    }),
+  );
 
   console.log('results: ', results);
 
   for (let i = 0; i < results.length; i++) {
-    if (!(results[i].reason instanceof Error)) continue;
+    if (results[i].status !== 'rejected') continue;
```
alukach, line 91: I reworked the logic here to what I believe is equivalent.

src/ml/modelInterfaces.ts
```diff
@@ -1,54 +1,65 @@
...
 
-const _getImage = async (image, config) => {
+const _getImage = async (image: ImageSchema, config: ModelInterfaceParams['config']) => {
   const url = 'http://' + buildImgUrl(image, config);
 
   try {
     const res = await fetch(url);
     const body = await res.arrayBuffer();
-    let imgBuffer = Buffer.from(body, 'binary');
+    let imgBuffer = Buffer.from(body);
```
alukach, line 14: I'm not 100% sure of this change. The typings for the `from` method are as follows:

```ts
from(
    arrayBuffer: WithImplicitCoercion<ArrayBuffer | SharedArrayBuffer>,
    byteOffset?: number,
    length?: number,
): Buffer;
/**
 * Creates a new Buffer using the passed {data}
 * @param data data to create a new Buffer
 */
from(data: Uint8Array | readonly number[]): Buffer;
from(data: WithImplicitCoercion<Uint8Array | readonly number[] | string>): Buffer;
/**
 * Creates a new Buffer containing the given JavaScript string {str}.
 * If provided, the {encoding} parameter identifies the character encoding.
 * If not provided, {encoding} defaults to 'utf8'.
 */
from(
    str:
        | WithImplicitCoercion<string>
        | {
            [Symbol.toPrimitive](hint: "string"): string;
        },
    encoding?: BufferEncoding,
): Buffer;
```

So, to pass in the `BufferEncoding` (e.g `'binary'`), the first argument should be a `str`, not an `ArrayBuffer`.  As such, I removed the `'buffer'` argument to comply with the required types.

nathanielrindlaub (approved):

PR #233, opened 2024-06-27
TypeScript: Convert tasks

alukach:
Convert tasks to TS.

---

part of #187


src/task/stats.ts
```diff
@@ -18,13 +20,12 @@ export default async function(task) {
...
   // stream in images from MongoDB
-  for await (const img of Image.aggregate(pipeline)) {
-
+  for await (const img of Image.aggregate<ImageSchema>(pipeline)) {
     // increment imageCount
     imageCount++;
 
     // increment reviewedCount
-    img.reviewed ? reviewed++ : notReviewed++;
+    isImageReviewed(img) ? reviewed++ : notReviewed++;
```
nathanielrindlaub, line 28: we don't need to re-assess whether the image is reviewed via `isImageReviewed(img)` here, as that will already be present on the image record. Change back to:

```
    img.reviewed ? reviewed++ : notReviewed++;
```

nathanielrindlaub (approved):
Looks good to me, but just for consistency's sake I'll run through these and use `import { type SomeTypeDef } ...` where appropriate in the imports before merging.

Issue #221, opened 2024-06-17
Make sure we're following Mongoose Typescript best practices


nathanielrindlaub:
There are a bunch of things I don't understand about how Typescript works with Mongoose, but we should read up the best practices and be sure we're implementing it correctly. Here are a few points of confusion:

1. **When to use the `HydratedDocument<>`generic?** From what I can tell from the [docs](https://mongoosejs.com/docs/typescript.html), HydratedDocument provides the document properties as defined in the Schema/document type, as well as all of the standard Mongoose properties/[methods](https://mongoosejs.com/docs/api/document.html). So my guess is any time we are using a Mongoose method like `SomeDocument.save()` we'll need to be dealing with a `HydratedDocument` type. If we're just returning the document without using any of it's methods we probably don't need to use `HydratedDocument`, but it probably wouldn't hurt, either. We obviously use `.save()` in lots of places, but I'm not sure what other Mongoose document methods we are using and where, but it would be worth auditing all of the model files to make sure we have full `HydratedDocument` coverage. Maybe we should just use it everywhere we create new documents or fetch them from the MongoDB, even if we're not using those methods? 
2. **ObjectId best practices?** I have a few questions here: (a) if we don't explicitly set an `_id` property in a Mongoose Schema and instead rely on Mongoose to add one of type `ObjectID` for us (e.g. like we do here [here](https://github.com/tnc-ca-geo/animl-api/blob/46939bf256102158c625cd27138e4df4e5990d0d/src/api/db/schemas/Project.ts#L55)), does `mongoose.InferRawDocType<SomeSchema>` know about the `_id` field, or do we have to explicitly set it in the schema? (b) Are we using `Schema.Types.ObjectId` in our schema definitions (as described in [here](https://mongoosejs.com/docs/typescript.html)) correctly? I know we use a lot of string `_ids`, for both good reasons and bad, but I did a search through the code for `Schema.Types.ObjectId` and only found one instance of it. Is it possible that we only ever explicitly set an id of type `ObjectId` once in the whole code base?
[...]

PR #236, opened 2024-07-25
Bug fixes for TypeScript refactor

nathanielrindlaub:
#187

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
         "seed-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
         "seed-db-dev": "STAGE=dev npm run seed-db",
         "seed-db-prod": "STAGE=prod npm run seed-db",
-        "export-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
+        "export-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
         "export-db-dev": "STAGE=dev npm run export-db",
         "export-db-prod": "STAGE=prod npm run export-db",
         "update-docs": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/updateDocuments.js",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
         "build": "tsc",
         "deploy-dev": "sls deploy --stage dev",
         "deploy-prod": "sls deploy --stage prod",
-        "seed-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
+        "seed-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
         "seed-db-dev": "STAGE=dev npm run seed-db",
         "seed-db-prod": "STAGE=prod npm run seed-db",
         "export-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
         "export-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
         "export-db-dev": "STAGE=dev npm run export-db",
         "export-db-prod": "STAGE=prod npm run export-db",
-        "update-docs": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/updateDocuments.js",
+        "update-docs": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/updateDocuments.js",
         "update-docs-dev": "STAGE=dev npm run update-docs",
         "update-docs-prod": "STAGE=prod npm run update-docs",
         "create-project": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/createProject.js",
--- a/src/@types/graphql.ts
+++ b/src/@types/graphql.ts
@@ -424,7 +424,7 @@ export type ImageAttempt = {
   _id: Scalars['ID']['output'];
   batch?: Maybe<Scalars['String']['output']>;
   created: Scalars['Date']['output'];
-  errs: Array<ImageError>;
+  errors: Array<ImageError>;
   metadata?: Maybe<ImageMetadata>;
   projectId: Scalars['String']['output'];
 };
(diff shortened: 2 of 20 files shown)


src/api/type-defs/objects/ImageAttempt.ts
```diff
@@ -24,6 +24,6 @@ export default /* GraphQL */ `
     batch: String
     created: Date!
     metadata: ImageMetadata
-    errors: [ImageError!]
+    errs: [ImageError!]
```
nathanielrindlaub, line 27: animl-ingest's `ingest-image` lambda [expects](https://github.com/tnc-ca-geo/animl-ingest/blob/00a0b2d6a30e0653abb284a88619aaadbcbb0faa/ingest-image/task.js#L185) there `errs` property to be there, so we should either make this a required property here and return an empty array if there are no `ImageErrors`, or update `ingest-image` to handle an undefined/null value.

nathanielrindlaub: Fixed.

src/api/db/models/Image.ts
```diff
@@ -181,14 +181,14 @@ export class ImageModel {
   static async createImage(
     input: gql.CreateImageInput,
     context: Pick<Context, 'user'>,
-  ): Promise<HydratedDocument<ImageAttemptSchema> & { errs?: ImageErrorSchema[] }> {
+  ): Promise<gql.ImageAttempt> {
```
nathanielrindlaub, line 184: It may be preferable to return GraphQL codegen-generated types in more model methods because (a) we don't need all the properties and methods that a Mongoose `HydratedDocument` contains and we also won't have to extend the `HydratedDocument`s so much with any additional properties that we tack on after fetching the record from the DB.

src/@types/graphql.ts
```diff
@@ -424,7 +424,7 @@ export type ImageAttempt = {
   _id: Scalars['ID']['output'];
   batch?: Maybe<Scalars['String']['output']>;
   created: Scalars['Date']['output'];
-  errors?: Maybe<Array<ImageError>>;
+  errs: Array<ImageError>;
```
jue-henry, line 427: Making sure this was a purposeful change:`errors` -> `errs`?

jue-henry: Ok definitely looks on purpose. Just curious about the reason

src/api/auth/authorization.ts
```diff
@@ -53,7 +52,7 @@ async function getUserInfo(req: APIGatewayProxyEvent, config: Config): Promise<U
 
     user['is_superuser'] = user['cognito:groups'].includes('animl_superuser');
     user['projects'] = projects;
-    user['curr_project_roles'] = projects[selectedProject] ? projects[selectedProject].roles : [];
+    user['curr_project_roles'] = selectedProject ? projects[selectedProject].roles : [];
```
jue-henry, line 55: I think this needs to be reverted, since projects might not have the selectedProject property

nathanielrindlaub: @jue-henry we do need to support calls with undefined `selectedProject`, because in at least one call (the initial `getProjects` call when the app is loading) the user doesn't yet have any Projects let alone a selected one. 

I wrote up an issue to figure out a better way to handle that here: https://github.com/tnc-ca-geo/animl-api/issues/237

nathanielrindlaub: I reverted it back to errors in this [commit](https://github.com/tnc-ca-geo/animl-api/pull/236/commits/be820d3a7b7da0c5a6a068bcc0a42e118e647282).

jue-henry: @nathanielrindlaub Are you going to resolve the issue in this PR? Currently as superuser who does not have permissions in any other projects, this throws a 500 error.

package.json
```diff
@@ -9,10 +9,10 @@
         "build": "tsc",
         "deploy-dev": "sls deploy --stage dev",
         "deploy-prod": "sls deploy --stage prod",
-        "seed-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
+        "seed-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/seedDb.js",
         "seed-db-dev": "STAGE=dev npm run seed-db",
         "seed-db-prod": "STAGE=prod npm run seed-db",
-        "export-db": "AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
+        "export-db": "tsc && AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/exportDb.js",
```
jue-henry, line 15: @nathanielrindlaub not entirely sure, but should we add `tsc &&` to line 18 as well?

jue-henry, line 15: @nathanielrindlaub Don't we need to also add `tsc &&` to the update-docs command since we also import from the build folder?

src/api/db/models/Camera.ts
```diff
@@ -48,7 +51,7 @@ export class CameraModel {
           const newCamera = new WirelessCamera({
             _id: input.cameraId,
             make: input.make,
-            projRegistrations: [{ projectId, active: true }],
+            projRegistrations: [{ _id: new ObjectId(), projectId, active: true }],
```
jue-henry, line 54: @nathanielrindlaub another question just for my sake, but what is this id used for here?

jue-henry (approved):
Changes look good to me and testing on staging seems fine

src/@types/graphql.ts
nathanielrindlaub: I did change it back to `errors` in the end.

src/api/db/models/Camera.ts
nathanielrindlaub: @jue-henry `ProjectRegistration`s are nested sub-documents in MongoDB parlance, so by default Mongoose was giving them an _id property without us explicitly setting it. To make the Mongoose type inference produce accurate TypeScript typings from our schemas, however, we had to start explicitly setting `_id` fields on all subdocuments (see ProjectRegistration schema [here](https://github.com/tnc-ca-geo/animl-api/blob/7504b98a01c14b0ae459ea01f6328dd2602ca5c4/src/api/db/schemas/WirelessCamera.ts#L4-L8)), which had a knock-on effect of needing to create the `_id`s manually when we create new records instead of letting Mongoose do it under the hood.

You can get more detail on that in this issue: https://github.com/tnc-ca-geo/animl-api/issues/221

package.json
nathanielrindlaub: good call. I'll do that now.

PR #271, opened 2024-10-10
Creating DeleteCamera Task

jue-henry:


src/api/db/models/Camera.ts
```diff
@@ -380,6 +381,61 @@ export class CameraModel {
...
+        context,
+      );
+    } catch (err) {
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  // NOTE: this method is called by the async task handler
+  static async deleteCamera(
```
jue-henry, line 407: Thinking about moving essentially all of these calls to the corresponding camera task handler, since it seems to be just a series of calls to different models

src/api/db/models/Camera.ts
```diff
@@ -380,6 +381,61 @@ export class CameraModel {
...
+        context,
+      );
+    } catch (err) {
+      if (err instanceof GraphQLError) throw err;
+      throw new InternalServerError(err as string);
+    }
+  }
+
+  // NOTE: this method is called by the async task handler
+  static async deleteCameraConfig(
```
jue-henry, line 407: Thinking about moving this to the camera task handler, since this is just a bunch of model operations

src/api/db/models/Project.ts
```diff
@@ -334,6 +367,49 @@ export class ProjectModel {
     }
   }
 
+  static async removeCameraFromViews(
```
jue-henry, line 370: Probably a bit convoluted way to do it, open to suggestions

src/api/db/models/Project.ts
```diff
@@ -334,6 +367,49 @@ export class ProjectModel {
...
+
+          console.log('deployments to be removed from views: ', projectDeps);
+
+          project.views.forEach((v, index) => {
+            // if view filters has the camera id or any of the deployment ids
+            if (
+              v.filters.cameras?.includes(input.cameraId) ||
+              projectDeps.some((d) => v.filters.deployments?.includes(d))
+            ) {
+              v.filters.cameras = v.filters.cameras?.filter((c) => c === input.cameraId);
```
nathanielrindlaub, line 395: I think we want a `filter((c) => c !== input.cameraId)` here, right?

src/api/db/models/Project.ts
nathanielrindlaub: I think this works! No ideas jumping to mind for streamlining it. I think all of those steps are necessary.

src/task/camera.ts
```diff
@@ -2,8 +2,53 @@ import { type User } from '../api/auth/authorization.js';
...
+    // Step 2: delete camera record from project
+    await ProjectModel.deleteCameraConfig(
+      {
+        cameraId: task.config.cameraId,
+      },
+      context,
+    );
+
+    // Step3: delete images associated with this camera
+    await DeleteImagesByFilter({
```
nathanielrindlaub, line 34: I haven't looked at the frontend code yet so I might be able to answer my own question, but it looks like we don't return the delete images task Id or do any polling here, so I'm assuming that the request from the frontend will trigger the `camera.deleteCameraTask()` mutation, the fronend will get a task Id back for this `DeleteCamera` task, perhaps do some polling for it, but meanwhile the `DeleteCamera` task is kicking off (and not awaiting the completion of) the separate `DeleteImagesByFilter` task, which may well not be completed by the time this `DeleteCamera` task is completed. Am I understanding that correctly?

If so we should think about adding polling here so we wait for the `DeleteImagesByFilter` task to complete before updating the parent `DeleteCamera` task as complete. Open to other ideas though. If we went that route, if there were errors added to the `DeleteImagesByFilter` task, I think we'd want to copy them into the `DeleteCamera` task too so that the frontend is aware of them.

src/task/camera.ts
```diff
@@ -2,8 +2,53 @@ import { type User } from '../api/auth/authorization.js';
...
+      projectId: task.projectId,
+      config: {
+        filters: {
+          cameras: [task.config.cameraId],
+        },
+      },
+      type: 'DeleteImagesByFilter',
+      user: task.user,
+    });
+    // Step 4: unregister camera
```
nathanielrindlaub, line 44: Just double-checking that this accounts for the fact that a Project can register a camera, later on unregister it, and a different Project can then register that same camera down the road. This allows a camera to be moved around to different preserves/projects etc.

At a glance I think `getWirelessCameras()` should return only the wireless cameras that _at some point_ had been registered to the current project, so in theory this condition will be met even if the wireless camera in question is now registered to a different project. 

I think that might(?) cause an issue in the `unregisterCamera()` step, because `unregisterCamera()` throws an error if the cam is not actively registered to the current project. 

I haven't tested this or dug too deep, but do you mind taking a look and testing?

nathanielrindlaub (commented):
Hey @jue-henry looking good! Few questions/comments here but it's very close.

src/api/db/models/Project.ts
jue-henry: You're right, didn't actually catch this because currently there's no way on the frontend to add a camera to the filters.

src/task/camera.ts
jue-henry: So, the way this is implemented is that that we don't actually create another task for the image deletion. I am reusing the code in the delete images task handler, but there aren't multiple tasks running at once. I do need to catch the errors with deletion and attach them to the task output here though

src/task/camera.ts
jue-henry: Yea you're right, the `unregisterCamera` method only works on active registrations. Will write another method for this

jue-henry: I guess how should the `projRegistration` be handled? Should it just be removed from the list of registrations and if it is the current active project, create a new one for `default_project`?

nathanielrindlaub: Yeah I think that's right. Find the `wirelesscamera`, if the current project is not the `active` one, just remove it from the `projRegistrations` array, but if it is active, we'll want to both remove it and set/create a `projResistration` for the `default_project`:

```javascript
// setting default project registration to active, copied from `CameraModel.unregisterCamera()` method:
const defaultProjReg = cam.projRegistrations.find((pr) => pr.projectId === 'default_project');
if (defaultProjReg) defaultProjReg.active = true;
else {
  cam.projRegistrations.push({
    _id: new ObjectId(),
    projectId: 'default_project',
    active: true,
  });
}
await cam.save();
```

src/api/db/models/Camera.ts
```diff
@@ -236,6 +237,72 @@ export class CameraModel {
...
+      if (activeReg?.projectId === projectId) {
+        const defaultProjReg = cam.projRegistrations.find(
+          (pr) => pr.projectId === 'default_project',
+        );
+        if (defaultProjReg) {
+          defaultProjReg.active = true;
+        } else {
+          // make sure there's a Project.cameraConfig record for this camera
+          // in the default_project and create one if not
+          defaultProj = await Project.findOne({ _id: 'default_project' });
```
jue-henry, line 270: @nathanielrindlaub slightly different approach to this than the `unregisterCamera` method above, but I make sure `default_project` exists first and then create a `CameraConfig` in the `default_project` project before creating the project registration. This way we make sure the project exists first, instead of having a camera project registration point to a non-existant project, but also not sure if this is even an issue that needs to be addressed.

nathanielrindlaub: @jue-henry I don't think there not being a `default_project` in the DB should be an issue. It gets [created](https://github.com/tnc-ca-geo/animl-api/blob/d0da6f81805e2721a1c2bc7e7df6d721c36f4f88/src/scripts/seedDb.js#L170C1-L184C5) when we initially seed the DB in the `src/scripts/seedDB.js` script. It is conceivable that someone might standing up their own Animl instance and not run the script, but that would cause a lot of other issues long before anyone every got to the point of deleting a camera, so I think it's safe to assume that `default_project` exists.

Dockerfile-test
```diff
@@ -7,7 +7,7 @@ COPY ./ $HOME/
 RUN apt-get update \
     && DEBIAN_FRONTEND=noninteractive apt-get install -y curl git
 
-RUN export NODEV='20.12.2' \
```
jue-henry, line 10: @nathanielrindlaub needed to upgrade for the docker tests to run. Merging for now but open to discussion

PR #276, opened 2024-10-21
Bulk Image Deletion Task

jue-henry:


src/task/image.ts
```diff
@@ -0,0 +1,9 @@
+import { type User } from '../api/auth/authorization.js';
+import { ImageModel } from '../api/db/models/Image.js';
+import { TaskInput } from '../api/db/models/Task.js';
+import type * as gql from '../@types/graphql.js';
+
+export async function DeleteImages(task: TaskInput<gql.DeleteImagesInput>) {
+  const context = { user: { is_superuser: true, curr_project: task.projectId } as User };
+  return await ImageModel.deleteImages(task.config, context);
```
jue-henry, line 8: Deciding if we should use the existing delete images function. Need to check why we are deleting images one at a time

nathanielrindlaub: One at a time as opposed to like a bulkWrite() or [deleteMany()](https://www.mongodb.com/docs/manual/reference/method/db.collection.deleteMany/) kind of operation?

I think it's because each "image" consists of:
- the three versions of the image file stored in S3
- the `image` record itself, as well as a corresponding `imageAttempt` and `imageError` record

So we go one at a time to remove all of those pieces for each image so if something goes wrong we don't have partial data for some of the images.

jue-henry: Yea makes sense after some thought. I think this will mean the async task will just end up calling original deleteImages

src/task/image.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  return { isOk: true };
+}
+
+export async function DeleteImages(task: TaskInput<gql.DeleteImagesInput>) {
+  const context = { user: { is_superuser: true, curr_project: task.projectId } as User };
+  while (task.config.imageIds?.length && task.config.imageIds.length > 0) {
+    const batch = task.config.imageIds?.splice(0, 100);
```
nathanielrindlaub, line 31: I don't know if this matters whatsoever but it feels a little weird to be mutating this input directly with `splice` like that. Maybe we clone the `task.config.imageIds` outside of the while loop and use that instead? Maybe I am just being insane? No strong feelings.

nathanielrindlaub (changes requested):
@jue-henry this is looking great to me just two small requests.

src/task/image.ts
jue-henry: Had similar idea, and i committed locally but didnt push haha. Yea we need to preserve the list of image ids to add it to the task output

PR #278, opened 2024-11-13
Optimize image deletion

jue-henry:
Final approach we landed on:

1. find image records on mongo
2. utilize mongo transactions to delete the appropriate records:
- we delete all image records, image attempt records, and image error records each in their own respective operation instead of 1 per image per type of record
- if an error occurs we abort all changes to mongo and return with an error

3. if mongo records are successfully deleted, we then delete the corresponding s3 objects all at once

Note: We opted to delete mongo records first, and then delete s3 objects, since a record without an image is more detrimental than a image without records. Thought was put into ensuring both are there, but converting the mongo deletion record operation to use `deleteMany` instead of `deleteOne` makes it so that we have far less per image errors available.


jue-henry:
Preliminary timing results after commit: 4c697c752946b080524ec61d25d1aa9bdd79c7a8

[log-events-viewer-result (1).csv](https://github.com/user-attachments/files/17740674/log-events-viewer-result.1.csv)

jue-henry:
Test results after commit 1fc07bb, moves all mongo record deletions to 1 call for each collecton:

[log-events-viewer-result (2).csv](https://github.com/user-attachments/files/17780715/log-events-viewer-result.2.csv)

Avg total deletion per 100 images: 0.877 seconds
Avg mongo deletion per 100 images:  0.1506 seconds
Avg s3 deletion per 100 images: 0.7264 seconds

src/api/db/models/Image.ts
```diff
@@ -143,16 +143,45 @@ export class ImageModel {
...
+        }),
+        await ImageError.deleteMany({
+          image: { $in: input.imageIds! },
+        }),
+      ]);
+      console.timeEnd('delete-images mongo records');
+
+      const keys: { Key: string }[] = [];
+      console.time('delete-images s3 records');
+      images!.forEach((image) => {
```
jue-henry, line 168: So kinda stuck on this bit. Should we only delete images that we can find mongo records for or should we just delete all images that were inputted by the user?

jue-henry:
Test results after commit [ec416c5](https://github.com/tnc-ca-geo/animl-api/pull/278/commits/ec416c57a24c2cde24299d76794940ecf0c93812), changing batch size to 300 and configuring errors better:

[log-events-viewer-result (3).csv](https://github.com/user-attachments/files/17821992/log-events-viewer-result.3.csv)

This test was done with 1k images, so it would be 3 batches of 300 and 1 batch of 100. The last batch was left out of the test, so we do not have a large sample size on this test
Avg total deletion per 300 images: 1.466 seconds
Avg mongo deletion per 300 images:  0.443 seconds
Avg s3 deletion per 300 images: 1.023 seconds

src/api/db/models/Image.ts
```diff
@@ -143,21 +143,80 @@ export class ImageModel {
...
+            // Commit the changes
+            await session.commitTransaction();
+          } else {
+            throw new Error('Failed to delete images');
+          }
+        } catch (error) {
+          // Rollback any changes made in the database
+          await session.abortTransaction();
+          await session.endSession();
+          throw new InternalServerError(error as string);
```
jue-henry, line 188: Aborts request and returns error if mongo records cannot be deleted

jue-henry:
Test results after commit [ec416c5](https://github.com/tnc-ca-geo/animl-api/pull/278/commits/ec416c57a24c2cde24299d76794940ecf0c93812), changing batch size to 300 and configuring errors better:

[log-events-viewer-result (4).csv](https://github.com/user-attachments/files/17940783/log-events-viewer-result.4.csv)

Total Time: ~11 seconds

This test was done with 4k images, so it would be 13 batches of 300 and 1 batch of 100. 

Avg total deletion per 300 images: 924.652 ms
Avg mongo deletion per 300 images: 136.349 ms
Avg s3 deletion per 300 images: 787.137 ms

jue-henry:
Test was done with ~52k images and delete by filter with a batch size of 300:

Total time: ~3 minutes

[log-events-viewer-result (6).csv](https://github.com/user-attachments/files/17982662/log-events-viewer-result.6.csv)

Avg total deletion per 300 images: 996.238 ms
Avg mongo deletion per 300 images: 115.175 ms
Avg s3 deletion per 300 images: 850.056 ms

package.json
```diff
@@ -55,7 +55,7 @@
         "mongoose": "^8.0.0",
         "prompt": "^1.2.0",
         "random-hex-color": "^1.0.1",
-        "sharp": "^0.33.4",
+        "sharp": "^0.33.5",
```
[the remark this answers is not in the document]
jue-henry: Binaries currently match, when dev opens back up I can test inference

src/api/db/models/Image.ts
```diff
@@ -144,23 +144,86 @@ export class ImageModel {
     context: Pick<Context, 'user'>,
   ): Promise<gql.StandardErrorPayload> {
     try {
-      const res = await Promise.allSettled(
-        input.imageIds!.map((imageId) => {
-          return this.deleteImage({ imageId }, context);
+      // Current limit of image deletion due to constraints of s3 deleteObjects
+      if (input.imageIds!.length > 300) {
```
jue-henry, line 148: Failing here because if not we would delete the mongo records but not s3 objects

jue-henry: Do we have a standard place for constants, looked in config and not sure it would be the same thing

package.json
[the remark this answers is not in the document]
jue-henry: reverted

PR #287, opened 2025-01-02
Label bulk deletion

lessej:
**Context**

Uses `bulkWrite` for label deletion operations to try and improve performance.

**TODO**
- [ ] write tests to verify that the operations are working as expected
- [ ] benchmark to figure out what the max should be

### DIFF
--- a/src/api/db/models/Image.ts
+++ b/src/api/db/models/Image.ts
@@ -1095,91 +1095,57 @@ export class ImageModel {
       'objects.labels.labelId': input.labelId,
     });
 
-    // Objects with single label which is input.labelId
-    const removableObjects = allImagesWithLabel.reduce((acc, img) => {
-      const matchingObjects = img.objects
-        .filter((obj) => obj.labels.length === 1 && obj.labels[0].labelId === input.labelId)
-        .map((obj) => obj._id);
-      return [...acc, ...matchingObjects];
-    }, [] as mongoose.Types.ObjectId[]);
-
-    // Locked objects with > 1 label where firstValidated is input.labelId
-    const unlockableObjects = allImagesWithLabel.reduce((acc, img) => {
-      const matchingObjects = img.objects
-        .filter((obj) => {
-          // If the object doesn't have > 1 labels or it is not locked, skip it
-          if (obj.labels.length <= 1 || obj.locked === true) {
-            return false;
-          } 
-          const firstValidated = obj.labels.find((lbl) => lbl.validation && lbl.validation.validated);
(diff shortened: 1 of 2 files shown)


src/api/db/models/Image.ts
```diff
@@ -1077,6 +1077,125 @@ export class ImageModel {
     );
   }
 
+  /**
+   * (n: n-m-q) Query for any images that have objects that have the label
```
jue-henry, line 1081: Is this formal set notation? For me this kind of reads like subtraction of m and q. Not sure if this can be a bit more clear

src/api/db/models/Image.ts
```diff
@@ -1077,6 +1077,125 @@ export class ImageModel {
...
+      });
+
+      return [...operations, ...removeOperation, ...unlockOperations, ...standardOperations];
+    }, []);
+
+    if (operations.length === 0) {
+      return true;
+    }
+
+    const res = await Image.bulkWrite(operations);
```
jue-henry, line 1195: So a very quick search mentions that theres a limit of 100k to bulk operations, not sure if we will end up hitting this. If that is a worry, then we can utilize mongo [transactions](https://mongoosejs.com/docs/transactions.html), so we can break up the bulk operations while keeping the atomicity, although we would lose some of the performance.

src/api/db/models/Image.ts
lessej: Sorry, this was just me making notes to myself before I started writing the code.  I will clean this up in a more readable way.

lessej:
In testing on dev, I was able to get up to around ~80k operations within the 30 second timeout.  This was tested using ~10,000 images each with 8 labels that needed to be removed.

nathanielrindlaub (approved):
@lessej this looks great and good to merge as far as I can tell. I think in the next iteration we will want to set an operation threshold above which we kick the work to the async task lambda, but unless you have anything else to add to this PR we can save all of that for a separate one.

src/api/db/models/Project.ts
```diff
@@ -795,22 +795,18 @@ export class ProjectModel {
       const label = project.labels?.find((p) => p._id.toString() === input._id.toString());
       if (!label) throw new DeleteLabelError('Label not found on project');
 
-      const count = await ImageModel.countImagesByLabel([input._id], context);
-
-      if (count > MAX_LABEL_DELETE) {
```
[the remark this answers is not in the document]
lessej: Removed in https://github.com/tnc-ca-geo/animl-api/pull/287/commits/8eae515b58449d17bd4c09d72220eec489ff85e6

PR #301, opened 2025-01-20
165: Filtering by tags

nathanielrindlaub:


nathanielrindlaub:
- [ ] check that tags can be added to Views
- [ ] and that when tags are deleted from a Project they also get removed from any views they were a part of

PR #322, opened 2025-06-04
add speciesnet-all support

geohacker:
Fixes https://github.com/tnc-ca-geo/animl-ml/issues/130

Once the ISO code field is added on the frontend, we need to wire that up. The necessary line is commented out for now.

### DIFF
--- a/src/ml/modelInterfaces.ts
+++ b/src/ml/modelInterfaces.ts
@@ -296,7 +296,8 @@ const speciesnet: InferenceFunction = async (params) => {
     image_data: imgBuffer.toString('base64'),
     bbox: speciesnetBbox,
     components: mode,
-    // ...(mode === 'all' && params.country && { country: params.country }) TODO: wire up after country code is added to params
+    ...(mode === 'all' && params.country && { country: params.country }),
+    ...(mode === 'all' && params.admin1Region && { admin1Region: params.admin1Region }),
   };
 
   // Choose the endpoint based on the mode. Default is realtime.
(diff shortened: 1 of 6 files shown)


src/ml/modelInterfaces.ts
```diff
@@ -276,22 +276,27 @@ const speciesnet: InferenceFunction = async (params) => {
...
+  }
+
   let bbox: BBox = label?.bbox ? label.bbox : [0, 0, 1, 1];
   const speciesnetBbox: BBox = label?.bbox ? _toSpeciesNetFormat(label.bbox) : [0, 0, 1, 1];
 
   const payload = {
     image_data: imgBuffer.toString('base64'),
     bbox: speciesnetBbox,
     components: mode,
+    // ...(mode === 'all' && params.country && { country: params.country }) TODO: wire up after country code is added to params
```
nathanielrindlaub, line 299: Per this [comment](https://github.com/tnc-ca-geo/animl-ml/issues/130#issuecomment-2933357784), let's also include `admin1_region`. See SpeciesNet [docs](https://github.com/google/cameratrapai?tab=readme-ov-file#input-format).

nathanielrindlaub (changes requested):

nathanielrindlaub:
@geohacker, this is looking good; I added the 2 new select fields to the Automation Rules form in [this frontend PR](https://github.com/tnc-ca-geo/animl-frontend/pull/326), so I think all that's left is enabling the `admin1_region` on the speciesnet endpoint side. 

One thing to note: the `AutomationRule.action.admin1Region`s get stored not as two-letter state codes, but in this hyphenated format that look like `US-CA`. I think that is the proper ISO 3166-2 formatting but elsewhere in the speciesnet docs they indicate that the API wants the two-letter state abbreviation. So that might take some testing/reformatting to get working.

geohacker:
Added admin1_region support to the ML api https://github.com/tnc-ca-geo/animl-ml/pull/131 and deployed

PR #323, opened 2025-06-04
Add Aggregate Objects at Image level for Annotations Export

sandrahoang686:


nathanielrindlaub (approved):
Looks great @sandrahoang686 !

PR #326, opened 2025-06-19
Burst stats

oliverroick:
Adds bursts stats to the API

This uses the `GetStats`, and to differentiate between image/object stats and bursts stats, we're introducing `AggregationLevel` to the `QueryStatsInput`.

### DIFF
--- a/src/task/getBursts.ts
+++ b/src/task/getBursts.ts
@@ -1,0 +1,107 @@
+import _ from 'lodash';
+import { DateTime } from 'luxon';
+
+import { ProjectModel } from '../api/db/models/Project.js';
+import { buildPipeline, idMatch } from '../api/db/models/utils.js';
+import { findRepresentativeLabel } from './utils.js';
+import { TaskInput } from '../api/db/models/Task.js';
+import { CameraConfigSchema, DeploymentSchema, FiltersSchema } from '../api/db/schemas/Project.js';
+import { AggregationLevel } from '../@types/graphql.js';
+import Image, { ImageSchema } from '../api/db/schemas/Image.js';
+
+
+const MAX_SEQUENCE_DELTA = 2;
+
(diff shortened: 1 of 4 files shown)


src/task/getBursts.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+  };
+
+  for (const dep of deployments) {
+    const depPipeline = structuredClone(pipeline);
+    depPipeline.push({
+      $match:{
+        deploymentId: dep._id,
+      }
+    });
```
nathanielrindlaub, line 72: As with the Individual Detections query, we'll want to stream in the images in chronological order, e.g.:

https://github.com/tnc-ca-geo/animl-api/blob/31f274f4532d564b24e1c7c0a48e80c10bedaded/src/scripts/analyzeMLSequenceLevel.js#L274

src/task/getBursts.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    let sequence: ImageSchema[] = [];
```
nathanielrindlaub, line 74: I know I haven't been entirely consistent, but can we try to use "burst" instead of "sequence" consistently across the code and UI? I've heard both used but we might as well pick one and stick with it.

src/task/getBursts.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+      }
+    }
+
+    for (const label of sequenceLabels) {
+      burstLabelList[label] = Object.prototype.hasOwnProperty.call(burstLabelList, label)
+        ? burstLabelList[label] + 1
+        : 1;
+    }
+  };
```
nathanielrindlaub, line 64: This logic looks right to me! I haven't done any testing yet though.

nathanielrindlaub (changes requested):

PR #328, opened 2025-07-10
fix: Alerts images should have proper expiration

alukach:
In 7841b55, we changed the unit of the buildImgUrl ttl from minutes to years. alerts.js was expecting it to be in minutes. This commit updates the value to be provided in years.


jue-henry:
resolved in other PR

PR #330, opened 2025-07-15
Add independence interval input param

oliverroick:
Adds new input parameter for the `GetStats` task allowing to specify the interval length for independent detections.


nathanielrindlaub (approved):
Looks great. Thanks @oliverroick!

PR #346, opened 2025-08-15
feat: add adjustable start and end for analysis window

lessej:
**Context**

Adjusts the given start and end date to avoid analyzing images that are from before or after the deployment window.

### DIFF
--- a/src/scripts/analysisConfig.js
+++ b/src/scripts/analysisConfig.js
@@ -7,6 +7,7 @@ const analysisConfig = {
   PROJECT_ID: 'sci_biosecurity',
   START_DATE: '2023-4-28',
   END_DATE: '2024-5-29',
+  ADJUSTABLE_WINDOW: true,
   ML_MODEL: 'mirav2', // first use of 'mirav2' was 2023-4-28
   TARGET_CLASSES: [
     // class naming convention: '<label.name>:<label._id>'
--- a/src/scripts/analyzeMLObjectLevel.js
+++ b/src/scripts/analyzeMLObjectLevel.js
@@ -37,7 +37,7 @@ import { ProjectModel } from '../../.build/api/db/models/Project.js';
  * STAGE=prod AWS_PROFILE=animl REGION=us-west-2 node ./src/scripts/analyzeMLObjectLevel.js
  */
 
-const { ANALYSIS_DIR, PROJECT_ID, START_DATE, END_DATE, ML_MODEL } = analysisConfig;
+const { ADJUSTABLE_WINDOW, ANALYSIS_DIR, PROJECT_ID, START_DATE, END_DATE, ML_MODEL } = analysisConfig;
 
 const TARGET_CLASSES = analysisConfig.TARGET_CLASSES.map((tc) => ({
   predicted_id: tc.predicted.split(':')[1],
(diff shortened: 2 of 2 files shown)


lessej:
@nathanielrindlaub took a stab at a very quick implementation of the start and end date fine-tuning.  Let me know if this is along the lines of what you are thinking.

lessej:
@nathanielrindlaub Updated this to use DB queries instead of looking through objects in memory.  Anecdotally, it does seem a little bit more performant and the code is much nicer.

We do tell the user that we're adjusting the date but the output file and other console logs will still use the original date.  I wasn't sure if we should use the adjusted window or if the user would still want to see the dates they set in their original config

src/scripts/analysisConfig.js
```diff
@@ -5,8 +5,9 @@
 const analysisConfig = {
   ANALYSIS_DIR: '/analysis',
   PROJECT_ID: 'sci_biosecurity',
-  START_DATE: '2023-4-28',
+  START_DATE: '2023-2-28',
```
lessej, line 8: This was just so I could check if it actually moved the dates or not.  Will revert

src/scripts/analyzeMLObjectLevel.js
```diff
@@ -138,17 +138,78 @@ function FVLValidatesPrediction(obj, tClass) {
...
 }
 
+async function tryAdjustAutomationWindow() {
+  console.log('attempting to adjust automation window...');
+  const imagesInAutomationWindow = await Image.aggregate([
+    {
+      $match: {
+        projectId: PROJECT_ID,
+        dateAdded: {
+          $gte: new Date(START_DATE),
```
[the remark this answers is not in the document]
nathanielrindlaub: Yeah honestly I think ideally we wouldn't constrict the query by `dateAdded` at all so as to look the model usage across the entire life of the Project.

src/scripts/analyzeMLObjectLevel.js
```diff
@@ -138,17 +138,78 @@ function FVLValidatesPrediction(obj, tClass) {
...
+            labels: {
+              $elemMatch: {
+                mlModel: ML_MODEL
+              }
+            }
+          }
+        }
+      },
+    },
+    { $sort: { dateAdded: 1 } }
```
nathanielrindlaub, line 163: I'm not sure if this would help at all, but instead of pulling all of the images into memory, you could do 2 separate queries, one with `{$sort: { dateAdded: 1 }` and one with `{$sort: { dateAdded: -1 }`, and add a `{$limit: 1 }` to the aggregations so you only get the oldest matching image and the newest, respectively. Again not sure if that would be any more performative but would be a bit less taxing on memory I suppose.

src/scripts/analyzeMLObjectLevel.js
```diff
@@ -138,17 +138,78 @@ function FVLValidatesPrediction(obj, tClass) {
...
+  ]);
+
+  const firstMlLabelAfterStart = imagesInAutomationWindow.shift();
+  const lastMlLabelAterStart = imagesInAutomationWindow.pop();
+
+  if (!firstMlLabelAfterStart || !lastMlLabelAterStart) {
+    throw new Error('unable to find a valid first and last image in automation window.');
+  }
+
+  const newStart = firstMlLabelAfterStart.dateAdded > (new Date(START_DATE))
```
nathanielrindlaub, line 173: We should probably +1 day to the start and -1 day to the end to make sure we're not including any partial days.

src/scripts/analysisConfig.js
```diff
@@ -7,6 +7,7 @@ const analysisConfig = {
   PROJECT_ID: 'sci_biosecurity',
   START_DATE: '2023-4-28',
   END_DATE: '2024-5-29',
+  ADJUSTABLE_WINDOW: true,
```
nathanielrindlaub, line 10: I would maybe call this something that's a little more descriptive like `AUTO_ADJUST_TIME_WINDOW`. Or just add a comment explaining what it does.

src/scripts/analyzeMLObjectLevel.js
```diff
@@ -138,17 +138,78 @@ function FVLValidatesPrediction(obj, tClass) {
   }
 }
 
+async function tryAdjustAutomationWindow() {
```
nathanielrindlaub, line 141: This seems solid for a first pass at a solution, and I don't want to over think it, but there are circumstances when users use model A for some period of time then try out model B then decide A was a better option so the switch back to A. This solution doesn't account for that of course but I think that's ok for now.

PR #348, opened 2025-08-28
Adding awaiting prediction logic back

jue-henry:
This reverts commit 1ac34076d41edb1f46f3e4f35f4aa38db1da9832.

### DIFF
--- a/src/api/db/models/Image.ts
+++ b/src/api/db/models/Image.ts
@@ -968,6 +968,7 @@ export class ImageModel {
           },
           { retries: 2 },
         );
+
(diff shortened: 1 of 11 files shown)


src/api/db/models/Image.ts
```diff
@@ -962,6 +962,7 @@ export class ImageModel {
 
             // set image as unreviewed due to new labels
             image.reviewed = false;
+            image.awaitingPrediction = false;
```
jue-henry, line 965: @nathanielrindlaub adding this here to save an additional db save

PR #350, opened 2025-09-01
feat: build temporary validation lists

lessej:
**Context**

Attempts to build a validation list for ml labels.  Takes an ML label, finds objects with that label, and builds a set of other labels on the objects.  Some things to think about:

- Is there a way to make it more efficient?  It's O(n^2) at least right now.  Can some of the in-memory operations be offloaded to the DB query?
- What to do with labels that a user has added but then removed -- it seems like these are still on the object

### DIFF
--- a/src/scripts/generateValidating.js
+++ b/src/scripts/generateValidating.js
@@ -1,0 +1,59 @@
+import { analysisConfig } from "./analysisConfig.js"
+import { getConfig } from '../../.build/config/config.js';
+import { connectToDatabase } from '../../.build/api/db/connect.js';
+import Image from '../../.build/api/db/schemas/Image.js';
+
+export const generateValidationList = async (analysisConfig, predictedLabels) => {
+  const { PROJECT_ID, START_DATE, END_DATE, ML_MODEL } = analysisConfig;
+
+  const config = await getConfig();
+  console.log('Connecting to db...');
+  await connectToDatabase(config);
+
+  const validatingLabels = predictedLabels.reduce((acc, lbl) => {
+    return { ...acc, [lbl]: [] }
+  }, {});
+
+  console.log('collecting images...');
+  const images = await Image.aggregate([{
+    $match: {
+      projectId: PROJECT_ID,
+      dateAdded: {
+        $gte: new Date(START_DATE),
+        $lte: new Date(END_DATE),
+      },
+      reviewed: true,
+    },
+  }]);
+
+  console.log('building validation lists...');
+  const validationLists = images.reduce((imgAcc, img) => {
+    img.objects.forEach((obj) => {
+      const mlLabel = obj.labels.find((lbl) => lbl.mlModel === ML_MODEL);
+      if (!mlLabel) {
+        return;
+      }
+
+      const validating = validatingLabels[mlLabel.labelId];
+
+      if (!validating) {
+        return;
+      }
+
+      const ids = obj.labels.reduce((acc, lbl) => {
+        if (lbl.mlModel === 'megadetector') {
+          return acc;
+        }
+        return acc.concat(lbl.labelId);
+      }, []);
+
+      validatingLabels[mlLabel.labelId] = new Set([...validating, ...ids]);
+    })
+    return imgAcc;
+  }, validatingLabels);
+
+  return validationLists;
+}
+
+// Example
+console.log(generateValidationList(analysisConfig, ['rodent', 'skunk']));


src/scripts/generateValidating.js
```diff
@@ -0,0 +1,170 @@
...
+    const projectDbTime = performance.now();
+    const project = await Project.findOne({
+      _id: PROJECT_ID
+    });
+    console.log(`Project db query time: ${(performance.now() - projectDbTime) / 1000} seconds`);
+
+    const projectLabels = project.labels.reduce((acc, lbl) => {
+      acc[lbl._id] = lbl.name;
+      return acc;
+    }, {});
```
nathanielrindlaub, line 31: When we run the `analyzeML.js` scripts, we're running them to analyze the performance of a single model, so the `analysisConfig.TARGET_CLASSES` only needs to include labels for the classes that that model predicts. So rather than using `SKIP_MODELS` in your config here it could just be `ML_MODEL`, and then we only add the `Project.labels` that are predicted by that model in our `TARGET_CLASS` list.

nathanielrindlaub: That's also why I used the key `predicted` in the `TARGET_CLASS` array. We should probably use "predictedLabels" here too rather than "projectLabels" where appropriate.

```javascript
...
    {
      predicted: 'rodent:rodent',
      validation: [
        'rodent:rodent',
        'mouse:mouse',
        'rat:rat',
        'vole:4a6973cd-5666-4277-b2eb-292a688e96c0',
      ],
    },
...
```

src/scripts/generateValidating.js
```diff
@@ -0,0 +1,170 @@
...
+          projectId: PROJECT_ID,
+          dateAdded: {
+            $gte: new Date(START_DATE),
+            $lte: new Date(END_DATE),
+          },
+          reviewed: true,
+          'objects.labels.labelId': {
+            $in: projectLabelIds
+          },
+          'objects.labels.validation.validated': true,
```
nathanielrindlaub, line 54: @lessej what's the thinking with this line ^?

src/scripts/generateValidating.js
```diff
@@ -0,0 +1,170 @@
...
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+      {
+        $set: {
```
nathanielrindlaub, line 84: Can you remind me what you're trying to do here with there `$set`ing of the `Image.filteredLabels` field?

src/scripts/generateValidating.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+    console.log(`Project db query time: ${(performance.now() - projectDbTime) / 1000} seconds`);
+
+    const projectLabels = project.labels.reduce((acc, lbl) => {
+      acc[lbl._id] = lbl.name;
+      return acc;
+    }, {});
+    const projectLabelIds = Object.keys(projectLabels);
+
+    // Prepare map { labelId: [validatingLabelId] }
```
nathanielrindlaub, line 34: This map should match the shape of the existing `TARGET_CLASS` array in the analysisConfig.js`, right? See example entry copied into the comment above.

src/scripts/generateValidating.js
```diff
@@ -0,0 +1,170 @@
...
+    const validationLists = images.reduce((imgAcc, img) => {
+      img.objects.forEach((obj) => {
+        if (!obj.firstValidLabel || obj.firstValidLabel.length < 1) {
+          return;
+        }
+
+        const labelIds = obj.filteredLabels.map((lbl) => lbl.labelId);
+        const firstValidLabelId = obj.firstValidLabel.shift().labelId;
+        const curr = validatingLabels[firstValidLabelId];
+        validatingLabels[firstValidLabelId] = new Set([...curr, ...labelIds]);
```
nathanielrindlaub, line 129: I'm having a little trouble understanding this. Seems a little odd to use a reducer here. You're both dropping in `validatingLabels` as an initial value (132) then referencing it and manipulating the out-of-scope variable from within the reducer loop? But then returning the `imageAcc` after each run of the reducer function (131)? Then I kind of assume is what the `reduce()` returns and assigns `imageAcc` to `validationLists` once it's iterated over all the images? Maybe I'm not tracking or something but I think this could be accomplished with a simple forEach loop:

```javascript

const validatingLabels = projectLabelIds.reduce((acc, lbl) => {
      return { ...acc, [lbl]: [] };
    }, {});

images.forEach((img) => {
      img.objects.forEach((obj) => {
        ...
         validatingLabels[firstValidLabelId] = new Set([...curr, ...labelIds]);
      }
}

File: src/scripts/generateValidating.js [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+
+    const validationLists = images.reduce((imgAcc, img) => {
+      img.objects.forEach((obj) => {
+        if (!obj.firstValidLabel || obj.firstValidLabel.length < 1) {
+          return;
+        }
+
+        const labelIds = obj.filteredLabels.map((lbl) => lbl.labelId);
+        const firstValidLabelId = obj.firstValidLabel.shift().labelId;
+        const curr = validatingLabels[firstValidLabelId];
```
nathanielrindlaub, line 128: I think this might be backwards. The `firstValidLabel` should be added to the array of `validating` labels, not used as the `predicted` label.

PR #351, opened 2025-09-04
Fixing internal label creation

jue-henry:


src/api/db/models/Image.ts
```diff
@@ -981,12 +979,14 @@ export class ImageModel {
           );
         }
       }
+      image.awaitingPrediction = false;
```
jue-henry, line 982: @nathanielrindlaub on second thought this seems like the more correct way of doing things since we want to save each label to the image individually and `awaitingPrediction` shouldn't really be false until all labels are saved

jue-henry: Alternatively we can make this change on the last label being added

PR #359, opened 2025-10-12
Add support for creation datetime offset in the Image data model

ejcunningham:
- Update the Image schema to allow setting an offset for the creation date in MS, `dateTimeOffsetMs`
- Add a mongoose virtual property, `dateTimeAdjusted`, which returns either the computed timestamp accounting for the offset or `dateTimeOriginal`
- Add graphql resolver and mutation functions for setting the offset of a single image (this may not actually be used in production, but I wanted to use it for testing/validation)
- Substitute usages of `dateTimeOriginal` with `dateTimeAdjusted` (behind the API only, still would need to update call sites in other repos)

Verified locally that the the graphql mutation function properly sets `dateTimeOffsetMs` in `dev` and that I can query for dateTimeAdjusted in cases where `dateTimeOffsetMs` is both set/unset

### DIFF
--- a/src/api/resolvers/Fields.ts
+++ b/src/api/resolvers/Fields.ts
@@ -16,6 +16,5 @@ export default {
           buildImgUrl(parent, context.config, size),
         ]),
       ),
-    dateTimeAdjusted: (parent: ImageSchema) => parent.dateTimeAdjusted,
   },
 };


src/api/resolvers/Fields.ts
```diff
@@ -16,5 +16,6 @@ export default {
           buildImgUrl(parent, context.config, size),
         ]),
       ),
+    dateTimeAdjusted: (parent: ImageSchema) => parent.dateTimeAdjusted,
```
nathanielrindlaub: @ejcunningham do we actually need this field-level resolver here if we are enriching the `Image` schema at the Mongoose level with the `dateTimeAdjusted` virtual? I genuinely am not sure, but I would have guessed that if we are shimming the `Image` records with a new field via the virtual, we wouldn't also need to do it in the GraphQL layer.

src/api/type-defs/inputs/SetTimestampOffsetInput.ts
```diff
@@ -0,0 +1,6 @@
+export default /* GraphQL */ `
+  input SetTimestampOffsetInput {
+    imageId: ID!
```
nathanielrindlaub: This looks great for now but just wanted to flag that in the future, in addition to supporting setting the offset on a single image, we'll need to be able to target multiple images.

You can take a look at how we approached this for [deleting images ](https://github.com/tnc-ca-geo/animl-api/blob/cf89ccb3832cca67f578a708c55b48c07411e036/src/api/type-defs/root/Mutation.ts#L4-L6) as a reference: essentially there is a `deleteImages` mutation that takes an [array of `imageIds` as an input](https://github.com/tnc-ca-geo/animl-api/blob/cf89ccb3832cca67f578a708c55b48c07411e036/src/api/type-defs/inputs/DeleteImagesInput.ts#L3C1-L3C20) for when users want to delete a small number of images at once and we can safely do it synchronously, then there's a separate mutation called `deleteImagesTask` which I think takes the same input (array of IDs), but kicks it to a async Task, and finally there's a `deleteImagesByFilterTask`, which accepts a set of filters as it's input and then querys and deletes all images that match those filters (as a Task).

I'm imagining we'll need to do something similar here too, but I'm getting ahead of myself.

Side note: I noticed the [`ImageModel.DeleteImages()`](https://github.com/tnc-ca-geo/animl-api/blob/cf89ccb3832cca67f578a708c55b48c07411e036/src/api/db/models/Image.ts#L151) method does use MongoDB transactions, but I think we only used them in that instance because we needed to delete records from three separate MongoDB collections (`Image` records, `ImageAttempt` records, and any `ImageErrors` records that matched the inputed IDs), and wanted to makes sure that if one failed we could abort the rest of the deletions in the transaction. So I'm not sure whether they are necessary for setting the timestamp offsets on lots of images, but IDK - I'd be curious to hear your thoughts.

PR #360, opened 2025-10-15
Checking for error types in ML handler

jue-henry:


src/ml/handler.ts
```diff
@@ -171,17 +172,19 @@ async function singleInference(config: Config, record: Record): Promise<void> {
     // Note: hacky JSON parsing below due to odd error objects created by graphql-request client
     // https://github.com/jasonkuhrt/graphql-request/issues/201
     const errParsed = JSON.parse(JSON.stringify(err));
-    const hasDuplicateLabelErrors = errParsed.response.errors.some(
```
jue-henry: I should have paid more attention to the comment above, it seems like the graphql library we are using doesn't return the correct types.

`console.log('error type: ', typeof err);  ==> object`
`console.log('error instanceof dupe: ', err instanceof DuplicateLabelError);  ==> false`
`console.log('error instanceof graph: ', err instanceof GraphQLError);  ==> false`

PR #365, opened 2025-10-23
Undo references to dateTimeAdjusted

ejcunningham:
We've identified performance problems with dateTimeAdjusted. Undo references to it until we can fix the performance issues

### DIFF
--- a/src/api/db/models/utils.ts
+++ b/src/api/db/models/utils.ts
@@ -287,19 +287,6 @@ export function buildPipeline(
     pipeline.push({ $match: isFilterValid(custom) });
   }
 
-  // Mongoose virtuals are not available in aggregations since they are computed in mongo
-  pipeline.push({
-    $addFields: {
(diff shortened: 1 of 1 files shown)


nathanielrindlaub:
@ejcunningham thanks for doing this. Mind also reverting the `$addFields` stage on the image aggregation pipeline? I just did some more testing with and without that stage and it appears to be adding a shocking amount of latency.

Project | getImages query duration | getImages query duration (w/o $addFields)
-- | -- | --
jldp | 23,112 ms | 26 ms
jldp | 12,750 ms | 20 ms
sci_biosecurity | 10,385 ms | 17 ms
sci_biosecurity | 4,066 ms | 18 ms
las_piletas | 18,404 ms | 22 ms
las_piletas | 12,136 ms | 13 ms
kilauea_point | 1,014 ms | 30 ms
kilauea_point | 248 ms | 10 ms

PR #377, opened 2025-12-31
Remap deployments in response to image offset changes

ejcunningham:
Updating timestamps for images may cause the image to be mapped to a different deployment. This PR does that for most offset requests. However, if the target deployment is in a different timezone it's unclear what the user would expect us to do. For now, this PR detects this case and returns an error.

Tested on dev that:
1. Timestamp changes take affect as expected when all images/deployments are in the same TZ
2. Timestamp changes are rejected when it might trigger an image to cross into a deployment of a different timezone

### DIFF
--- a/src/api/errors.ts
+++ b/src/api/errors.ts
@@ -137,3 +137,14 @@ export class DeleteCameraError extends GraphQLError {
     });
   }
 }
+
+export class TimestampOffsetValidationError extends GraphQLError {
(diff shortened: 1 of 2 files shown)


nathanielrindlaub (approved):
@ejcunningham this is looking great. I just did a bunch of testing on dev and everything seems to be working as expected, so I'm going to merge and deploy.

